PR #44, opened 2013-10-30
List recipes in a way that includes overlapping names

jessepeterson:
I've found that I have enough remote autopkg repos to have overlapping names. When searching for something using `autopkg list-recipes` I only see the names and not which specific repository it comes from. In the case of duplicates (e.g. multiple repos having the same-named package) it leads to a false sense of which recipes exist. This patch all recipes and which repo directory they're listed in (or if they're an override). Welcome to suggestions to change.

### DIFF
--- a/Code/autopkg
+++ b/Code/autopkg
@@ -786,7 +786,7 @@ def list_recipes(argv):
 
     override_dirs = options.override_dirs or get_override_dirs()
     search_dirs = options.search_dirs or get_search_dirs()
-    recipes = set()
+    recipes = []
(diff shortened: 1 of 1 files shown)


n8felton:
It looks like #134 and #135 solved this PR? @jessepeterson can you confirm what you were looking for is now implemented and perhaps we can close this PR?

jessepeterson:
@n8felton Yes it appears that with these add'l switches my need is well fulfilled. Thanks for checking!

PR #67, opened 2014-01-21
Update Copier.py to only overwrite files when given overwrite key

hansen-m:


hansen-m:
@timsutton Any updates on if this might be added into a future release? FWIW I've been running with this modification for over a year without any issues (but I also don't use any of the .jss or .munki recipes).

hansen-m:
Bumping this issue... I'm still making this change manually on my instance of AutoPkg and haven't run into any issues. Would it make sense to clone Copier.py with the change listed and name it something like SafeCopier.py?

nmcspadden:
This is an ancient PR, but I think it's still valid and relevant. Seems like a branch was created for this 5 years ago:
https://github.com/autopkg/autopkg/tree/copier-overwrite-default

I updated this on my fork to match current standards:
https://github.com/autopkg/autopkg/compare/master...nmcspadden:copier-overwrite-default?expand=1

nmcspadden:
If you have no objections, I'm going to close this PR and create a new one based on my updated version. You are also welcome to take it if you want the contributor credit for it.

PR #102, opened 2014-07-21
Improve PathDeleter to allow process to continue if a target file don't exist

ygini:
Improvement useful for deployment scenario and don't break existing recipe.
The new feature is enable only when delete_path_silent_error_for_unexisting_path is true.
Log feature has been updated to be more specific.


tcinbis:
That would be a great addition, because maybe the files does not exists so everything is fine. Example deleting _MASReceipt for AppStore Apps.

PR #143, opened 2015-01-19
Update URLDownloader.py to handle urls with spaces

vmule:


gregneagle:
Why don't you url-encode the URL before passing it to URLDownloader? Typically URLs are not passed around with spaces within. Can you provide a non-contrived example where this is needed?

Code/autopkglib/URLDownloader.py
```diff
@@ -125,7 +126,8 @@ def main(self):
         # Download URL.
         url_handle = None
         try:
-            request = urllib2.Request(url=self.env["url"])
+            url = urllib.quote(self.env["url"], 'http://')
```
gregneagle: The second parameter to urllib.quote is a list of "safe" characters; that is, a list of characters to NOT encode. Not sure what you are aiming for here...

vmule:
Can I encode the URL within a recipe?

Here also an example:

URLTextSearcher: Found matching text (match): Wireshark 1.12.3 Intel 64.dmg
URLDownloader
Couldn't download http://wiresharkdownloads.riverbed.com/wireshark/osx/Wireshark 1.12.3 Intel 64.dmg: HTTP Error 400: Bad Request
Failed.

Code/autopkglib/URLDownloader.py
vmule: Yes, indeed.

If I omit that the colon ':' will be encoded too:

And I would end up like this:

Couldn't download http://wiresharkdownloads.riverbed.com/wireshark/osx/Wireshark 1.12.3 Intel 64.dmg: unknown url type: http%3A//wiresharkdownloads.riverbed.com/wireshark/osx/Wireshark%201.12.3%20Intel%2064.dmg
Failed.

vmule: for some reason, I can't specify just the colon, i.e.:

url = urllib.quote(self.env["url"], ':')

I does not work either.

gregneagle: You'd want, at a minimum url = urllib.quote(self.env["url"], ':/') but then it would do the wrong thing if some crazy path had a colon in it.

Again, lets see a real-world example of why it's not acceptable to require that the URL provided to URLDownloder be properly encoded from the start.

gregneagle:
Where are you getting a download URL of the format " http://wiresharkdownloads.riverbed.com/wireshark/osx/Wireshark 1.12.3 Intel 64.dmg" ? Whatever is generating that should url encode the pathname _before_ assembling it into a url. That's the correct way to generate properly-encoded URLs. Passing a possibly-already-encoded URL to urllib.quote can lead to undesirable results.

gregneagle:
Specific example: start with a properly-encoded URL:

foo = "http://wiresharkdownloads.riverbed.com/wireshark/osx/Wireshark%201.12.3%20Intel%2064.dmg"

Now run it through your proposal

urllib.quote(foo, "http://") (or the more correct url lib.quote(foo, ":/")

You get:

'http://wiresharkdownloads.riverbed.com/wireshark/osx/Wireshark%25201.12.3%2520Intel%252064.dmg'

It's been doubly-encoded; the percent signs get encoded again. Most likely this URL will now fail.

vmule:
I'm using this 2 processors in a recipe

```
         <dict>
              <key>Processor</key>
              <string>URLTextSearcher</string>
              <key>Arguments</key>
              <dict>
                  <key>url</key>
                  <string>https://www.wireshark.org/download.html</string>
                  <key>re_pattern</key>
                  <string>Wireshark.*Intel.*64.dmg</string>
              </dict>
          </dict>
          <dict>
              <key>Processor</key>
              <string>URLDownloader</string>
              <key>Arguments</key>
              <dict>
                  <key>url</key>
                  <string>http://wiresharkdownloads.riverbed.com/wireshark/osx/%match%</string>
                  <key>filename</key>
                  <string>%NAME%.dmg</string>
              </dict>
          </dict>
```

If there is something I am missing, on how to encode the url before passing it to URLDownloader processor please advise, I'd happy to use that.
Thanks

gregneagle:
You might need to write a product-specific URL provider. The partial URL returned from this page needs to be URL-encoded before being assembled into a full URL. I don't think there's an existing provider to so that, and as I've shown, it's dangerous to pass an already encoded URL through urllib.quote, so you can't just do it indiscriminately; you have to add more intelligence.

vmule:
So you are saying that the autolib such URLTextSearcher.py and URLDownloader.py do not encode the urls at all, but you are concerned an encoded url might be passed to them? how? manually from a receipts or from a custom provider?
I think it may be worth to manage al this edge case and have a provider that will work in both cases, if possible.

gregneagle:
URLTextSearcher does a regex match against text downloaded from a URL. It's pretty generic and makes no assumptions about what you will do with the matched text; you might perform any arbitrary action with the text in subsequent recipe steps.

URLDownloader expects a proper URL.

You've found an interesting edge case, but your proposed fix will break properly encoded URLs. I can't be in favor of a change that will break things that currently work.

I suggest bringing this topic up on autopkg-discuss.

vmule:
we could do something like this:


>>> foo = "http://wiresharkdownloads.riverbed.com/wireshark/osx/Wireshark%201.12.3%20Intel%2064.dmg"
>>> foo2 = urllib.unquote(foo)
>>> foo2
'http://wiresharkdownloads.riverbed.com/wireshark/osx/Wireshark 1.12.3 Intel 64.dmg'
>>> urllib.quote(foo2, 'http://')
'http://wiresharkdownloads.riverbed.com/wireshark/osx/Wireshark%201.12.3%20Intel%2064.dmg'


not sure if it's a good thing, but it would solve the concern you had above.

Thanks

timsutton:
I feel like URLDownloader should be only handling a valid URL and not further transforming them (for example, `curl` won't automatically encode a URL with spaces), but perhaps we instead add an input variable to URLTextSearcher that would tell it to do an unquote/quote.

Normally it would not be set, but the recipe could set it if it's attempting to "fix" a page that's not listing a properly-quoted URL. I imagine this is pretty rare since I think this may be the first documented case.

vmule:
created a new pull request with the syggested fix.
I will revert the other changes to URLDownloader.py
Thanks

arubdesu:
Sorry as I'm only skimming the thread, but what's going on with Wireshark that @[jleggat](https://github.com/autopkg/jleggat-recipes/blob/master/Wireshark/Wireshark.download.recipe) or @[cgerke](https://github.com/autopkg/cgerke-recipes/blob/master/Wireshark/Wireshark.download.recipe) aren't running into the same issue?

vmule:
This is weird, looking again at @cgerke recipe, he is using https for the download.
If I use that, even with spaces, the url is reached and it's working.
Maybe a urllib2 bug when using http?

timsutton:
Also, the cgerke recipe takes the dmg matched by URLTextSearcher and subs it into a new download URL.

If I try a non-https url reached by cgerke's recipe using `curl`, "http://2.na.dl.wireshark.org/osx/Wireshark 1.12.3 Intel 64.dmg`, it redirects to an https URL and doesn't return the dmg, it returns a download page. If I `curl` the https version, it doesn't redirect and starts downloading.

So it's also possible this is all some subtlety with how their webserver/website is configured, and if the issue is actually caused by how any http urls redirect to https and how this interacts with activity when not using a web browser.

vmule:
I see the same behaviur, I think we can close this!
Thanks

vmule:
I'm having the same issue with this url too :(

http://c758482.r82.cf2.rackcdn.com/Sublime Text Build 3065.dmg

gregneagle:
Reopening, but the proposed fix is still unacceptable as we can't blindly urlquote a string that may or may not already be urlquoted (IOW, urlquoting is not idempotent)

vmule:
Please look at this:
https://github.com/vmule/autopkg/compare/patch-2
This allow the possibility to specify when you want to urllib.quote a url.
I'll create the pull request if you think is acceptable.
Cheers

timsutton:
Sure, this seems ok to add to URLTextSearcher for edge cases.

Since you only ever need to reference `url_quote = self.env.get('url_quote', False)` once, you may as just put this statement at L119 directly, resulting in less code spread around the processor. If you want to open a PR with this change we can merge.

An aside:
I was pretty sure someone else already had a working ST3 recipe, and I was right: https://github.com/autopkg/keeleysam-recipes/blob/master/SublimeText/SublimeText3.download.recipe

It works because it uses the Sparkle processor to find the latest update, and the Sparkle processor automatically does url-quoting on enclosure URLs. (Sparkle itself [does this](https://github.com/sparkle-project/Sparkle/blob/89081ca030c0de218400f7c0f97530df524d687d/Sparkle/SUAppcastItem.m#L131-L134)).

So again, there is a solution for your specific case. I assume we will eventually run into this again, though.

gregneagle:
One concern I have is that the URLTextSearcher can return multiple key/value pairs and the proposed patch would either url quote none of them or all of them. Hope we don't run into a situation where that's bad.

My argument would be that if you need to find a URL to download, make sure you find a valid one!

timsutton:
Ugh, yes. So it gets uglier, but perhaps doable: `url_quote` could instead be renamed to something like `url_quote_match_variables` and be an array (None or empty by default) of keys to quote. In probably 99% of cases it would be a one-element array.

So if one is doing a simple RE with URLTextSearcher and storing the match in a `match` output variable, `url_quote_match_variables` contains a `match` string:

```plist
<dict>
    <key>Processor</key>
    <string>URLTextSearcher</string>
    <key>Arguments</key>
    <dict>
        <key>url</key>
        <string>%SUPPORT_CENTER_URL%</string>
        <key>url_quote_match_variables</key>
        <array>
            <string>match</string>
        </array>
        <key>re_pattern</key>
        <string>download\.asp\?(asset=centrify-suite[^"]*)</string>
    </dict>
</dict>
```

Otherwise `url_quote_match_variables` might contain only `url`, etc.

jessepeterson:
Could this be solved with a simple URLStringEncoder Processor that just, well, URL encodes a string?

gregneagle:
If the URL discovered/matched/otherwise retrieved needs to be URL encoded/quoted this will address the issue until the vendor changes their page/feed.

But since vendors sometimes make other changes that result in recipes needing to be updated, this should be no _worse_.

jessepeterson:
[Here's](http://stackoverflow.com/questions/14074731/are-iris-valid-as-html-attribute-values) an interesting StackOverflow on whether non-URL encoded values are allowed in HTML attributes. Short answer: HTML4: no, HTML5 sorta: supports IRIs but with some extra conditions.

I mention this in the context of modifying URLTextSearcher: I like its existing generic approach however I'd have to be honest with the fact that 99% of the time it's being used for an HTML document to find a download URL. Given URLs may (now) be specified validly with non-ASCII characters.

Perhaps *some* processor (not sure if URLTextSearcher, URLDownloader, or another new processor) could implement a check: if a URL already contains a '%' character then assume it is already URL encoded and use it verbatim, else URL encode? Or check for commonly-escaped characters like spaces and encode them? I wonder how browsers handle this somewhat ambiguous case.

PR #144, opened 2015-01-19
Added optional url_quote to the input in URLTextSearcher.py

vmule:
Added optional url_quote to the input in URLTextSearcher.py, if true will encode the match, if false nothing will be done.
As suggested in pull request 143
https://github.com/autopkg/autopkg/pull/143#issuecomment-70543364

### DIFF
--- a/Code/autopkglib/URLTextSearcher.py
+++ b/Code/autopkglib/URLTextSearcher.py
@@ -16,6 +16,7 @@
 """See docstring for URLTextSearcher class"""
 
 import re
+import urllib
 import urllib2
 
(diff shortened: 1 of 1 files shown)


vmule:
Do you think is still meaninful?
Thanks

timsutton:
I was thinking about it, and I'm not sure. Your case was the first time this has ever come up, and there are hundreds of download recipes out there. I'm not really inclined to add more code to core processors for which there currently isn't any use, even if it's a small change.

But, if this does come up again, I'm happy to take another look at adding it.

vmule:
thanks for the answer, and this very useful tool :)

PR #145, opened 2015-02-03
Add "find" as alias of "search" in verbs

bochoven:
I keep writing find instead of search, it also makes more sense grammatically.

### DIFF
--- a/Code/autopkg
+++ b/Code/autopkg
@@ -1406,6 +1406,9 @@ def main(argv):
     """Main routine"""
     # define our subcommands ('verbs')
     subcommands = {
+        "find": {
+            "function": search_recipes,
+            "help": "Search for recipes on GitHub."},
         "help": {
             "function": display_help,
             "help": "Display this help"},


timsutton:
I'm not totally sure I like the idea of aliasing verbs in AutoPkg, in the interest of keeping things simpler and more explicit (one vs. multiple ways to do things).

There's also already been some prior user confusion around `search` vs. `list-recipes`, although it's hopefully resolved when one reads the short descriptions printed using `autopkg help`. Adding another alias to that, I'm not sure people wouldn't instead think it's analogous to running a `find` command (to find data local to the machine).

To me, "search" is analogous to doing a search on the web (or in this case, it's literally a GitHub code search). Docker uses `docker search` as the verb for searching for images in the public repositories, for example.

Anyway, that's my logic - does that help things make more sense?

bochoven:
I don't mind if you close this in favour of simplicity, it's just a minor inconvenience.

timsutton:
Sorry. You _might_ be able to hack together something clever with shell functions in an rc file...

PR #147, opened 2015-02-09
Make run_recipes() except block handle failures more broadly.

sheagcraig:
# Exception handling during multiple recipe runs
When running a single recipe, a recipe failure is either handled by the individual failing processor and passed up as a ProcessorError and AutoPkg reports a failure, or AutoPkg stops with an exception. Either way, one recipe failed, and you get immediate information about it.

However, when running multiple recipes, an unhandled exception halts AutoPkg execution, rather than continuing with the next recipe. This *should* only be a problem with custom processors, but in practice, and as non standard-library processors (included in recipe repos?) proliferate, those which don't look for exceptions and in-turn raise ```ProcessorError``` have the potential to halt multiple-recipe runs.

This isn't a huge problem, but it does prevent the final reporting from happening. When running a long list of recipes that could take a significant period of time, if one recipe failed, I want to know what *else* happened: which new products were downloaded, uploaded to Munki, etc, as well as the failure.

One could argue that *any* error should halt AutoPkg if it isn't handled. I know that personally I want an individual recipe error to *not* stop execution, but I guess that depends on how severe the consequences would be for different organizations. (I don't see any...)

The wiki could include some information about making sure to wrap failure-prone sections with a try/except that percolates a ProcessorError up, but that seems to be beyond the scope of the standard expectations of an AutoPkg user or recipe-author.

It would be nice if the individual processors all handle this, but in practice, some higher level insurance seems appropriate. 

# CTRL-C Handling
It appears that a lot of the std-lib processors are checking for ```BaseException```, which is the base class for *all* exceptions, including the CTRL-C/```KeyboardInterrupt``` exception. Therefore, during a multiple-recipe ```autopkg run```, some processors will just fail and continue with the next recipe, whereas those without the 
```
except BaseException as e:
	raise ProcessorError()
```
idiom will stop AutoPkg execution. I think it's safe to assume that CTRL-C should make everything stop, no matter what.

If you'd like me to go through stdlib processors and switch them to use ```Exception``` instead of ```BaseException``` to make CTRL-C handling consistent, let me know and I'll update the PR.

### DIFF
--- a/Code/autopkg
+++ b/Code/autopkg
@@ -258,20 +258,20 @@ def load_recipe(name, override_dirs, recipe_dirs, preprocessors=None, postproces
                 log_err('Could not find parent recipe for %s' % name)
         else:
             recipe["RECIPE_PATH"] = recipe_file
-    
+
     if recipe and preprocessors:
         steps = []
         for preprocessor_name in preprocessors:
             steps.append({"Processor": preprocessor_name})
         steps.extend(recipe["Process"])
         recipe["Process"] = steps
-    
+
     if recipe and postprocessors:
         steps = recipe["Process"]
         for postprocessor_name in postprocessors:
             steps.append({"Processor": postprocessor_name})
         recipe["Process"] = steps
-    
+
     return recipe
 
 
--- a/Code/autopkg
+++ b/Code/autopkg
@@ -27,6 +27,7 @@ import re
 import shutil
 import subprocess
 import time
+import traceback
 import autopkglib.github
 
 from urlparse import urlparse
@@ -1291,6 +1292,7 @@ def run_recipes(argv):
             log_err("Failed.")
             failure["recipe"] = recipe_path
             failure["message"] = unicode(err)
+            failure["traceback"] = traceback.format_exc()
             failures.append(failure)
             autopackager.results.append({'RecipeError': unicode(err).rstrip()})
 
--- a/Code/autopkglib/__init__.py
+++ b/Code/autopkglib/__init__.py
@@ -18,7 +18,6 @@
 
 import os
 import sys
-
 import imp
 import FoundationPlist
 import pprint
@@ -286,7 +285,7 @@ class Processor(object):
                 self.env[variable] = flags["default"]
                 self.output("No value supplied for %s, setting default value "
                             "of: %s" % (variable, self.env[variable]),
-                           verbose_level=2)
+                            verbose_level=2)
             # Make sure all required arguments have been supplied.
             if flags.get("required") and (variable not in self.env):
                 raise ProcessorError(
@@ -468,15 +467,15 @@ class AutoPackager(object):
                 print >> sys.stderr, unicode(err)
                 raise AutoPackagerError(
                     "Error in %s: Processor: %s: Error: %s"
-                    %(identifier, step["Processor"], unicode(err)))
+                    % (identifier, step["Processor"], unicode(err)))
 
             output_dict = {}
             for key in processor.output_variables.keys():
                 # Safety workaround for Processors that may output
-                # differently-named output variables than are given in their
-                # output_variables
-                # TODO: develop a generic solution for processors that can
-                #       dynamically set their output_variables
+                # differently-named output variables than are given in
+                # their output_variables
+                # TODO: develop a generic solution for processors that
+                #       can dynamically set their output_variables
                 if processor.env.get(key):
                     output_dict[key] = self.env[key]
             if self.verbose > 1:
@@ -534,17 +533,17 @@ def add_processor(name, processor_object):
 
 #pylint: disable=invalid-name
 def extract_processor_name_with_recipe_identifier(processor_name):
-    '''Returns a tuple of (processor_name, identifier), given a Processor name.
-    This is to handle a processor name that may include a recipe identifier, in
-    the format:
+    '''Returns a tuple of (processor_name, identifier), given a Processor
+    name.  This is to handle a processor name that may include a recipe
+    identifier, in the format:
 
     com.github.autopkg.recipes.somerecipe/ProcessorName
 
     identifier will be None if one was not extracted.'''
     identifier, delim, processor_name = processor_name.partition('/')
     if not delim:
-        # if no '/' was found, the first item in the tuple will be the full
-        # string, the processor name
+        # if no '/' was found, the first item in the tuple will be the
+        # full string, the processor name
         processor_name = identifier
         identifier = None
     return (processor_name, identifier)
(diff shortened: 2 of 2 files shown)


timsutton:
For the added safety of catching unhandled exceptions, we might be able to implement this in autopkglib "closer" to where handling `ProcessorError` is done:

https://github.com/autopkg/autopkg/blob/master/Code/autopkglib/__init__.py#L465-L471

I think it might also be useful to be able to log a traceback in these cases, since they're by definition unexpected. We could potentially use `sys.exc_info` and [`traceback`](https://docs.python.org/2/library/traceback.html) to get these and put this info in the error.

As for the disparity between Exception classes in the various core processors, maybe we address these separately. I agree that `BaseException` would be too broad if we want to make sure at least a `KeyboardInterrupt` aborts a run.

sheagcraig:
I moved the ```except Exception:``` to __init__.py.

I'll put in a PR for the std_lib processors separately.

sheagcraig:
FWIW, there's also a few \t converted to four spaces in one of the commits. Didn't want to send any surprise riders in!

timsutton:
We could handle these unexpected exceptions in _addition_ to the existing AutoPackagerError except: block. This way we can log the "expected" ProcessorErrors first, then fall back to the unexpected ones in a second block. This would be also be needed to be able to handle those differently, if we could get some traceback info for example.

gregneagle:
Still leaves open the possibility of not catching an unexpected custom exception that inherits from Exception.

sheagcraig:
Is it more important to avoid pylint warnings or to handle custom exceptions that inherit from Exception? Can a pylint hint just be added in? Or are you trying to avoid those as much as possible?

timsutton:
Interesting that PyLint would complain - I don't really see anything in the Exception Python docs that discourages deriving from Exception.

It might be ok to just move the custom exceptions down one level to inherit from StandardError then. I don't know whether that'd be creating more problems, but maybe worth trying.

sheagcraig:
I think Greg is pointing out the possibility for non-AutoPkg exceptions derived from Exception. Like in Python std library modules. Or custom Exceptions, like in JSSImporter.

sheagcraig:
Just looking through the 2.7 lib, the exceptions are almost all inheriting from Exception rather than StandardError.

gregneagle:
As are exceptions in FoundationPlist and most anything else I've written that have custom exceptions.

sheagcraig:
```grep -R 'class .*(Exception):' /System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/```

**Comment by gregneagle** [comment]
Perhaps _selectively_ suppressing PyLint warnings and doing "except Exception" only when absolutely needed is the better approach.

**Comment by gregneagle** [comment]
"Interesting that PyLint would complain - I don't really see anything in the Exception Python docs that discourages deriving from Exception."

PyLint complains when you use "except Exception" as catching a too-general exception. It does not complain about  "except BaseException", but it's obvious that it _should_, since that's even worse, really.

**Comment by sheagcraig** [comment]
So...

- Have the process loop catch ```ProcessorError```, than ```Exception``` (with a pylint hint).
- And the std processors should catch more specific exceptions per task rather than ```Exception``` or ```BaseException```.

**Comment by gregneagle** [comment]
"Is it more important to avoid pylint warnings or to handle custom exceptions that inherit from Exception? "

PyLint is trying to guide us towards Python best practices, so when PyLint complains, we need to understand the "why" and not just turn off the warnings.

**Comment by sheagcraig** [comment]
Putting only one broad ```except Exception:``` in the process section, with documentation explaining how it is a kind of safety valve for the processors, points to recognizing that best practice that pylint is pointing too.

The processors in question catch ```BaseException```, which does seem strange in that it avoids pylint complaining that it's too broad, especially considering if it were "restricted" to ```Exception```, pylint would *then* warn that it was too broad.

There are around 9 broad uses in the std processors. I can go through them and find more appropriate, more limited except: blocks. For example, Copier has a try/except block that catches ```BaseException```. The inside of the block, however, is all ```shutil``` functions: copyfile, copytree, and copy. All of these raise an ```shutil.Error```, which is subclassed from EnvironmentError. This may be a more appropriate thing to catch, and let the processor try/except catch the more unexpected errors.

Other processors might just have one try/except wrapping the entire processor code, which would need to be broken down into smaller, more specific try/except blocks around projected points of failure. 

This puts the responsibility for handling those unpredictable errors on the ```AutoPackager.process``` rather than on individual processors. It also improves the self-documentation of the code to have specific, targeted try/except blocks, with an overall safety valve at the highest level.

**Comment by sheagcraig** [comment]
I dropped back to the last upstream commit and redid this.

@timsutton At one point I had a multiple try/except/except, i.e.

```
try:
	processor.process()
except ProcessorError as err:
	# ...
except Exception as err:
	# ...
```
But the handling ended up being the same, so I settled on catching ```Exception``` and documenting the reason why it was so broad.

Also, I added the traceback to the failure dictionary. I tried adding it into the final reporting, but as you can imagine, the traceback is potentially a lot of mess to add in. I tried it several ways; indenting it another level, etc, and it always seemed wrong. So at the moment it just gets stored. I can address outputting that (possibly with a higher verbosity level?) with the reporting work I'm doing on another branch.

You have to grab the traceback while it is being handled. In py3 the exception itself has a ```__traceback__``` property, but py2's ```sys.last_traceback``` gets deleted after handling occurs.

**Comment by timsutton** [comment]
I think we've needed something like this for a while, and while it's related to #148 this is a simpler change so I'd like to merge this first. I think it makes sense to be storing the traceback in `failures` as you did. If I intentionally introduce an unhandled exception in URLDownloader, a run report plist looks like:

```plist
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0">
<dict>
	<key>failures</key>
	<array>
		<dict>
			<key>message</key>
			<string>Error in com.github.autopkg.download.Cyberduck: Processor: URLDownloader: Error: global name 'foo' is not defined</string>
			<key>recipe</key>
			<string>Cyberduck.download</string>
			<key>traceback</key>
			<string>Traceback (most recent call last):
  File "./autopkg", line 1459, in run_recipes
    autopackager.process(recipe)
  File "/Users/tsutton/git/github/autopkg/Code/autopkglib/__init__.py", line 479, in process
    % (identifier, step["Processor"], unicode(err)))
AutoPackagerError: Error in com.github.autopkg.download.Cyberduck: Processor: URLDownloader: Error: global name 'foo' is not defined
</string>
		</dict>
	</array>
	<key>summary_results</key>
	<dict/>
</dict>
</plist>
```
..so the `message` item remains brief but the traceback is there if desired. Exceptions that were handled using ProcessorError work similarly.

PR #148, opened 2015-02-10
Stdize catching on Exception  to allow KeyboardInterrupt through.

sheagcraig:
`except BaseException:` and a bare `except:` catch `KeyboardInterrupt`. The changes below replace processors with either of these except clauses with `except Exception` to standardize the handling of CTRL-C/`KeyboardInterrupt`. This is meant to complement work done in autopkg/autopkg#147, allowing CTRL-C to halt execution no matter what.

Also, pymode caught a blank space at the end of a line in `InstallFromDMG.py`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,5 @@
+### 0.4.3 (Unreleased)
+
 ### 0.4.2 (December 12, 2014)
 
 ADDITIONS:
--- a/Code/autopkglib/AppDmgVersioner.py
+++ b/Code/autopkglib/AppDmgVersioner.py
@@ -92,10 +92,10 @@ class AppDmgVersioner(DmgMounter):
             try:
                 self.env["bundleid"] = info["CFBundleIdentifier"]
                 self.env["version"] = info["CFBundleShortVersionString"]
-                self.output("BundleID: %s" % self.env["bundleid"])
-                self.output("Version: %s" % self.env["version"])
-            except Exception as err:
+            except KeyError as err:
                 raise ProcessorError(err)
+            self.output("BundleID: %s" % self.env["bundleid"])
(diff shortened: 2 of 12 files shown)


gregneagle:
PyLint complains a lot if you try to catch the "too-general" exception "Exception". I obviously misunderstood the documentation on Python exceptions I read a while back when I changed all these to BaseException -- that's even more general.

I'd propose using "StandardError". According to here: https://docs.python.org/2/library/exceptions.html

exception StandardError
The base class for all built-in exceptions except StopIteration, GeneratorExit, KeyboardInterrupt and SystemExit. StandardError itself is derived from Exception.

This seems general enough to catch most exceptions without blocking KeyboardInterrupt. We'd need to test to see if PyLint complains.

timsutton:
I'd propose the same - I don't think if we've ever had a case of any of the Warning subclasses being raised (see the table at the bottom of that link from the Py docs).

gregneagle:
Sure, but it leaves out all of the non-built-in exceptions that inherit from Exception...

sheagcraig:
Going through just AppDmgVersioner, I discovered several things that are beyond the scope of this PR. Should I start making a bunch of independent PR's?
- Switch plist handling code out for FoundationPlist import
- Add plist_version_key input variable to match Versioner.
- Exception handling limited to KeyError in relevant section.

timsutton:
Agreed those are outside the scope. If you want to open more separate PRs for those that's fine, but we should probably deal with the main issue first.

sheagcraig:
Okay-went through and chose more appropriate and more narrow exceptions for all of the processors which were doing ```except BaseException:```.

sheagcraig:
Just reviewing everything I have open.
This now has conflicts; but more importantly, do I need to break these into per-processor PRs?

PR #150, opened 2015-02-14
Fix munkiimport

bochoven:
Munkiimport was returning 'duplicate pkg' if the first pkgid was found
in the pkgdb.


gregneagle:
In looking at the function you modified, i think I found some more logic errors. I think I've addressed them here: https://github.com/autopkg/autopkg/commit/f6c55e3f14b8ddeddd4038c4b39a4e97472e17e4

bochoven:
Resolved with f6c55e3

PR #155, opened 2015-02-23
Refactor AppDmgVersioner to use FoundationPlist. Add version key.

sheagcraig:
Add plist_version_key input variable to match other processor's ability
to use different version keys (defaults to CFBundleShortVersionString).


sheagcraig:
Just checking in-is this useful?

PR #160, opened 2015-03-11
Add `repo-log` and `repo-update --check` features

eahrold:
This PR adds the ability to view a simple commit history natively in autopkgr using `repo-log`.

```
$ autopkg repo-log --help
Usage: autopkg repo-log recipe_repo_path_or_url [...]
Show commit history of recipe repos.
You may also use 'all' to check all installed recipe repos.

Options:
  -h, --help       show this help message and exit
  --new            Show only unmerged commits to the recipe repo.
  --diff           Include file diffs for commits.
  --number=NUMBER  Number of commits to show.
```

There are a few issues which need ironed out.

1) **Paging.** Currently this uses `subprocess.Popen` which will spit out everything at once, which for older repos can be quite a lot. If switched to `subprocess.call()` git log paging is preserved, but if running 'all' repos chosen you would need to 'q' out of a git log for each repo. We could set this up to page unless 'all' is selected. Suggestions?

2) **Pretty format.** What I've put together is pretty minimal but I think registers nicely. Feel free to make any change.

3) **repo-update --check** I've added a simple `--check` argument to `repo-update` that compares the SHA-1 of the last commit on the remote server with that of the local repo and informs the user if there are updates available. It's nice in that it doesn't need to do a full `git fetch` to detect changes, but may make more sense for this to be synonymous with `repo-log --new` instead and redirect there.

Let me know if you have any suggestions or recommendations.
Thanks.

### DIFF
--- a/Code/autopkg
+++ b/Code/autopkg
@@ -661,11 +661,8 @@ def repo_update(argv):
                      "You may also use 'all' to update all installed recipe "
                      "repos." % ("%prog", verb))
 
-    parser.add_option("--check", action="store_true", default=False,
-                      help=("Check for new commits to the specified recipe repo."))
-
     # Parse arguments
-    options, arguments = parser.parse_args(argv[2:])
+    arguments = parser.parse_args(argv[2:])[1]
     if len(arguments) < 1:
         log_err("Need at least one recipe repo path or URL!")
         return -1
@@ -686,23 +683,9 @@ def repo_update(argv):
                 repo_dirs.append(repo_path)
 
     for repo_dir in repo_dirs:
+        log("Attempting git pull for %s..." % (message, repo_dir))
         try:
-            if options.check:
-                git_rem_sha1 = ['ls-remote', '--heads', 'origin', './.', 'master']
-                git_local_sha1 = ['rev-parse', 'master']
-                rem_sha1 = run_git(git_rem_sha1, git_directory=repo_dir)
-                loc_sha1 = run_git(git_local_sha1, git_directory=repo_dir)
-                if not rem_sha1.split()[0].strip() == loc_sha1.strip():
-                    print 'Repo updates available for %s' % os.path.basename(repo_dir)
-                    print 'To see changes run "%s repo-log --new %s"' % \
-                                (os.path.realpath(sys.argv[0]), repo_dir)
-            else:
-                log("Attempting git pull for %s..." % (message, repo_dir))
-                output = run_git('pull', git_directory=repo_dir)
-                if output:
-                    log(output)
-                else:
-                    log("%s is up-to-date..." % (repo_dir))
+            log(run_git(['pull'], git_directory=repo_dir))
         except GitError, err:
             log_err(err)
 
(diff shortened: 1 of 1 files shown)


timsutton:
Thanks! Sorry for the long delay. Let's discuss possible use cases again. The whole "issue/PR" workflow of GitHub means the discussion gets split in a couple places. Oh well:

Here are a few scenarios I can imagine:

1. AutoPkg user at the CLI says "Hey, what's new in the repos in my list?" Is there some way for AutoPkg to be able to summarize these changes reasonably?
1. AutoPkgr or some wrapper tool wants to be able to present some dialog to the user about whether there are updates, and if so, what they are.
1. A user wants to see exactly what changed in a specific repo, either new changes they haven't yet merged or what was recently merged. (Without the user needing to go change to `~/L/RecipeRepos/etc` and run Git commands there)

It's impossibly hard to generalize this enough - to make output detailed enough when you want it and condensed enough to avoid screens and screens of noise. But here are my immediate thoughts about these three hypothetical use cases:

1. There's a fair amount of whitespace in the summarized commit log, which is useful for looking at a specific repo, but it's hard to separate repos when doing `repo-log all`. I wonder if `repo-log all` instead triggers some variant of the `--oneline` option for each repo log, so that scanning the output quickly makes it easier to visualize.
1. Would AutoPkgr actually use something like `repo-log` for functionality like this, or would it resort to just doing its own Git parsing anyway? We always get output to stdout regardless of whether or not there is anything in `repo-log --new`, so it's not really something that could be used from a script or other process. It's probably too complicated to consider supporting its output as though it's some kind of API, however, because way more tools can just manipulate using Git directly.
1. This works well enough, probably because it already works similar to a default Git log.


As for your other points:

1)
Let's not bother worrying about paging. People can use the `--number` option to limit the number of commits. (I think something like `--last` might be a more intuitive option name..)

2)
A few formatting tweaks I found were helpful for the summary display, patch below: 1) adding `--color` to the patch output resulted in getting colored diffs when the `--diff` option would be used, 2) adding `--name-status` to the default output mode resulted in also listing the files changed, and  3) removing the redundant 'Changes' text next to every datestamp.

```diff
diff --git a/Code/autopkg b/Code/autopkg
index e8dacc9..eaa9c21 100755
--- a/Code/autopkg
+++ b/Code/autopkg
@@ -755,13 +755,15 @@ def repo_log(argv):
 
 
     if options.diff:
-        git_command.append('--patch')
+        git_command.extend(['--patch', '--color'])
+    else:
+        git_command.append('--name-status')
 
     if options.number > 0:
         git_command.append('-n%d' % options.number)
 
     for repo_dir in repo_dirs:
-        pretty_format = '--pretty=format:%CredChanges (%aD):%n  %Creset%s%n'
+        pretty_format = '--pretty=format:%Cred%aD:%n  %Creset%s%n'
         git_command.append(pretty_format)
         try:
             run_git(['fetch'], git_directory=repo_dir)
```

3)
I'm still less in favor of having these actions available under two different commands, just to keep complexity lower. The `ls-remote` command does make things faster, so one could still potentially use it within `repo-log --new` (as a check for whether a fetch is needed), but I'm not sure the difference is very noticeable for repos as small as these.

Just to help discoverability, I'd think it's fine if we want to add a note in the `repo-update` command help that `repo-log` exists.

A discussion I was having with Greg about this pointed to another larger issue. Many times, commit logs by themselves aren't that informative for the user. We update repos because things change and break over time. It would be even _more_ useful if when someone runs a recipe, if it fails, AutoPkg can then tell you if that recipe's repo has an update for the recipe file you ran which failed. It may be possible to later extend `repo-log`'s functionality to be able to be called by AutoPkg when a recipe fails if certain criteria satisfy in the Git histories.

eahrold:
I just reverted the repo-update to it's original implementation which should squash those bugs.
Also fixed up the formatting using a oneline-ish format when `all` is specified, as well as adapted your other suggestions.

I do like the idea of using this to cross checked when recipes fail, and possibly prompt to update the repo , it feels like a natural step with the interactive direction AutoPkg is heading.

I noticed the PR went out-of-square with master. 
Do you want me to either rebase, merge master into this branch, or just leave as is?

eahrold:
💀

PR #164, opened 2015-03-13
Add xz to valid tar extensions

bochoven:
To accommodate the Opera download format .

### DIFF
--- a/Code/autopkglib/Unarchiver.py
+++ b/Code/autopkglib/Unarchiver.py
@@ -28,7 +28,7 @@ EXTNS = {
     'zip': ['zip'],
     'tar_gzip': ['tar.gz', 'tgz'],
     'tar_bzip2': ['tar.bz2', 'tbz'],
-    'tar_xz': ['xz'],
+    'tar_xz': ['tar.xz', 'xz'],
     'tar': ['tar']
 }
 
--- a/Code/autopkglib/Unarchiver.py
+++ b/Code/autopkglib/Unarchiver.py
@@ -28,6 +28,7 @@ EXTNS = {
     'zip': ['zip'],
     'tar_gzip': ['tar.gz', 'tgz'],
     'tar_bzip2': ['tar.bz2', 'tbz'],
+    'tar_xz': ['xz'],
     'tar': ['tar']
 }
 


Code/autopkglib/Unarchiver.py
```diff
@@ -28,6 +28,7 @@
     'zip': ['zip'],
     'tar_gzip': ['tar.gz', 'tgz'],
     'tar_bzip2': ['tar.bz2', 'tbz'],
+    'tar_xz': ['xz'],
```
gregneagle: Shouldn't this be ['tar.xz', xz'] ? The original reports from thenikola referred to "tar.xz" and "tar.gz" extensions so it's not entirely clear.

gregneagle:
Should I be concerned that thenikola says that tar doesn't properly decompress these on 10.6.8 and he wasn't sure about anything other than 10.10.1?

bochoven:
I haven't tested this on 10.6. Opera is supported from 10.7 and up. I'm not sure if they rely on the Mac OS built in tools to unarchive the updates.

gregneagle:
It would be nice for _someone_ to test on 10.7 at least; if it works there (and we know/think it works on 10.10, we can pretty safely assume it works on the OSes in between)

timsutton:
A plain `tar -xf` doesn't work:

```
tar -xf /Users/test/Library/AutoPkg/Cache/com.github.bochoven.recipes.Opera.download/downloads/Opera_28.0.1750.40_Autoupdate.tar.xz
tar: Error opening archive: Child process exited with status 254Child process exited with status 254
```

But perhaps some other combinations of commands does..

timsutton:
To clarify, that was 10.7. 10.6 also gives a different error: `tar: Unrecognized archive format: Inappropriate file type or format`.

If the archive really is of an [LZMA format](http://en.wikipedia.org/wiki/Xz) then perhaps there's a different utility altogether that's standard on OS X that can deal with these?

timsutton:
So two possible solutions:

1) include an xz extract utility in your recipe and write a custom decompressor processor that calls it
2) we look at adding support for a "MinimumOSVersion"-like key in recipes that could be used judiciously for cases like this

I think whenever possible it's good for a recipe author to strive to support the same range of OSes supported by AutoPkg in recipes that are for public consumption.

bochoven:
Hmm I'm already starting to regret this recipe author thing..

The xz utility from http://tukaani.org/xz/ runs from 10.5 till 10.10, but I suppose you don't want to bundle that with the core autopkg tools?

gregneagle:
Nope.

bochoven:
Ok, closing this

timsutton:
FWIW, here's how @nmcspadden solved this:

https://github.com/autopkg/nmcspadden-recipes/tree/26748d7a45050e1ed79f7cf508e135be7cfe66b1/Minecraft

bochoven:
Thanks @timsutton that looks promising

PR #166, opened 2015-03-23
Return the version number from SparkleUpdateInfoProvider

antifuchs:
For my use case, I have to find the version number for the flat pkg contained in a versioned .dmg file linked from an appcast feed. Since the Sparkle provider already extracts the most reasonable version number possible, I'd like to have it available in my recipes. I hope this seems reasonable!


antifuchs:
OK, *this*, I'm pretty confident, will work (-:

timsutton:
Thanks!

PR #169, opened 2015-04-01
add try/except blocks for OS X specific imports for use on Windows/Linux

jgstew:
This is a step on the way to making AutoPkg run cross platform, which may never happen. AutoPkg does work just fine for automating the preparation of Windows software for deployment, it just requires an OS X machine to run on.

These changes should in no way affect AutoPkg working on OS X.

TODO:  code would need to be written to abstract the use of Foundation & CoreFoundation on non-OSX platforms.  (NSArray, NSDictionary, and CFPreferences in Code/autopkglib/__init__.py)

The processors with OS X dependencies like AppDmgVersioner, MunkiInstallsItemsCreator, and MunkiSetDefaultCatalog would not be useful on other platforms anyway, so they would not need to be altered.

StopProcessingIf would need altered if used cross platform.

### DIFF
--- a/Code/FoundationPlist/__init__.py
+++ b/Code/FoundationPlist/__init__.py
@@ -1,2 +1,6 @@
 # We're effectively using this package as module
-from FoundationPlist import *
+try:
+    from FoundationPlist import *
+except:
+    print "WARNING: using 'from plistlib import *' instead of 'from FoundationPlist import *' in " + __name__
+    from plistlib import *
--- a/Code/autopkg
+++ b/Code/autopkg
@@ -16,9 +16,15 @@
 """autopkg tool. Runs autopkg recipes and also handles other
 related tasks"""
 
+import sys
+
+if 'darwin' != sys.platform:
+    print "----------------------------------------------------------------------------------"
+    print "--  WARNING: AutoPkg is not completely functional on platforms other than OS X  --"
+    print "----------------------------------------------------------------------------------"
+
 import difflib
 import glob
-import sys
 import os
 import operator
 import optparse
--- a/Code/autopkg
+++ b/Code/autopkg
@@ -22,7 +22,13 @@ import sys
 import os
 import operator
 import optparse
-import FoundationPlist
+
+try:
+    import FoundationPlist
+except:
+    print 'WARNING: importing plistlib as FoundationPlist'
+    import plistlib as FoundationPlist
+
 import pprint
 import re
 import shutil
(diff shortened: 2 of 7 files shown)


timsutton:
This looks ok, if nothing else for the fact that it's also a useful way to make AutoPkg fail with some more helpful output in cases where someone has overwritten the Apple's system Python with another distribution.

Do you run AutoPkg on Windows by just passing in all of your search dirs, etc. using command switches?

jgstew:
I'm not actually running AutoPkg on Windows yet. I'm just interested in seeing of the OS X dependencies can be removed from AutoPkg. As of now AutoPkg will not run on the whole under Linux/Windows due to the use of CFPref parts of Foundation and CoreFoundation. The 'autopkg search' functionality does work just fine on Windows/Linux with this change.

timsutton:
A couple more comments I'll add directly in the commit..

jgstew:
I removed the unneeded "pass" statements, and added a message when running autopkg on non-OS X systems.

timsutton:
Sorry for letting this linger. Since it's harder for me to test at the moment in Windows, what happens when AutoPkg tries to read preferences? At first glance it seems like there would be exceptions raised from the missing imports for CFPreferences.

jgstew:
Yes, it does raise exceptions due to the missing imports for CFPrefs if you use AutoPkg in a way that uses those calls.

Either equivalent replacement library functions need to be written for Win/Linux, or those sections need to be abstracted to make different calls depending on the platform. 

To be clear, this pull request does not make AutoPkg functional cross platform. It just makes it so it will run with clear warnings about the potential issues. This is just a stepping stone to eventually make AutoPkg work cross platform, but a lot more work would need to be done to make that a reality.

The GIT recipe searching does work cross platform currently as it is pure python.

timsutton:
That makes sense. I thought that perhaps you were using processors you'd use for download recipes, but when I saw how many times `get_pref()` is used throughout the code for search directories and such.. it might make sense to include in your warning message a list of functions that are actually of use on non-OS X, or just 'search' if that's the only one currently.

timsutton:
Not sure if this change will help in terms of readiness for your PSUMAC session next week (and it still wouldn't be in a "released" version), but we wanted to merge this anyway, since it's a first step.

jgstew:
Thanks. This is definitely worth a mention for the PSU MacAdmins session, but it will definitely take some time before AutoPkg is cross platform. It is definitely something I hope to do more work towards at some point.

PR #171, opened 2015-04-06
Suppresses URL check in SparkleUpdateInfoProvider if PKG provided

nmcspadden:
SparkleUpdateInfoProvider.py will always check the Sparkle feed URL, even if a local PKG is specified with -p.  This tells Sparkle to ignore and not check the Sparkle feed URL if a PKG option is specified (which will also prevent the recipe from failing if internet access is unavailable and a local package is provided).


timsutton:
More cautiousness added here: 047fbb2ad777b63c221caeae99b6056716f676ce

PR #174, opened 2015-04-16
Handle NSPropertyListSerialization not returning either an error or data...

sheagcraig:
....


Code/FoundationPlist/FoundationPlist.py
```diff
@@ -98,7 +98,7 @@ def _plistToData(plistObject):
         (data, error) = (
             NSPropertyListSerialization.dataFromPropertyList_format_errorDescription_(
                 plistObject, NSPropertyListXMLFormat_v1_0, None))
-    if error:
+    if error or data is None:
         raise NSPropertyListSerializationException(error)
```
gregneagle: If error is None and data is None, the raised exception would be NSPropertyListSerializationException(None) -- are we sure that's what we want?

PR #176, opened 2015-04-20
Added support for launchd 2.0 socket retrieval.

MagerValp:
The API used by Code/autopkgserver/launch.py were deprecated in 10.10 and a much simpler one was added (see man 3 launch). I added launch2.py using the new API and a version check in autopkgserver to call the correct one.

I don't have a 10.9 machine set up for regression testing so I've only tested the new codepath for 10.10.


timsutton:
Thanks Per! Tested on older OSes and the old code path predictably still works.

PR #172, opened 2015-04-07
Added support for opening editor on override

nmcspadden:
When creating an override, the autopkg admin has to navigate to the override file manually to edit it. This specifies an -o option on autopkg make-override that allows specifying the path to an executable that will immediately open the override file, similar to what munkiimport does.

PR #181, opened 2015-05-08
Removing trailing period from make-override path output

n8felton:
Currently there is a trailing period in the output of the 'make-override' command that makes it a pain to quickly copy and paste the path from the output.

Example: http://i.imgur.com/XwMh3kD.gifv

This PR removes that trailing period to making copying and pasting this path easier; especially when making multiple override files back to back for a new setup.


timsutton:
Makes sense to me. This will be less necessary when we look at merging #172, but no reason to not make it easier.

Issue #126, opened 2014-10-20
Support Content-Encoding: gzip in URLDownloader


jessepeterson:
As mentioned in autopkg/jessepeterson-recipes#5 I have a package download that is getting served with a `Content-Encoding: gzip`. It appears urllib2 doesn't natively support this. So: would folks be averse to merging a PR to implement that functionality in autopkg? I could, I suppose, also implement logic on the recipe and/or recipe Provider side to fixup the download.

The place where this is an actual issue is when the Unarchiver processor tries to run ditto to decompress a file labelled `.zip` but is actually encoded as a `.gz`.

jessepeterson:
As a short term workaround how does this sound: Making the URLDownloader merely set a Content-Encoding output variable based on the corresponding HTTP header? In this way another recipe Processor step could take decisive action (or no action) based on what the content is encoded as.

I bring this up because the original recipe I had to work around this with *reverted* back to being non-gzip encoded. I'd guess perhaps they have differently configured load balancers or something: one that compresses and one that doesn't. Might be a crap shoot on what the download comes as so dynamically handling this would be valuable.

timsutton:
Since we've got all the headers from the response already, we could just shove _all_ of them in an output variable dict like `response_headers`. Then if for some edge cases like yours, you have data to for a custom processor to make a decision about.

Having played around a bit with that, though:

```patch
diff --git a/Code/autopkglib/URLDownloader.py b/Code/autopkglib/URLDownloader.py
index 3aefe39..dad1fa8 100755
--- a/Code/autopkglib/URLDownloader.py
+++ b/Code/autopkglib/URLDownloader.py
@@ -90,6 +90,11 @@ class URLDownloader(Processor):
                 ("Boolean indicating if the download has changed since the "
                  "last time it was downloaded."),
         },
+        "response_headers": {
+            "description":
+                ("A dictionary containing all headers from the server "
+                 "response, keyed by header name."),
+        },
     }

     def main(self):
@@ -172,6 +177,7 @@ class URLDownloader(Processor):
                     self.output("Using existing %s" % pathname)
                     return

+            self.env["response_headers"] = url_handle.info().headers
             # Download file.
             self.env["download_changed"] = True
             with open(pathname, "wb") as file_handle:
```

...you wouldn't be guaranteed the output variable exists - the processor will exit early if it's deemed unmodified. Even moving up to the initial try: catch: (looking for a 304 response) wouldn't get you access to any useful data in the `url_handle`.

[...]

PR #184, opened 2015-06-02
Urldownloader content encoding

jessepeterson:
PR for autopkg/autopkg#126. Feedback welcome!

### DIFF
--- a/Code/autopkglib/URLDownloader.py
+++ b/Code/autopkglib/URLDownloader.py
@@ -18,6 +18,7 @@
 import os.path
 import urllib2
 import xattr
+import zlib
 
 from autopkglib import Processor, ProcessorError
 try:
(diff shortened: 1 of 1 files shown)


timsutton:
Looks ok to me on first glance - are your Extensis recipes still an example of ones which would benefit from supporting gzip Content-Encodings in URLDownloader?

jessepeterson:
Yes, the UniversalTypeClient4 is. Today, anyway. :)

timsutton:
I've merged (with a couple small nitpicks) and tested it my usual run of about 50-60 recipes. Anyway, as far as impact to all other recipes, if the servers aren't sending back `Content-Encoding` response headers, then the gzip code paths aren't executed.

I'm curious why you set `gzip_handle = zlib.decompressobj(16 + zlib.MAX_WBITS)`? How'd you arrive at that magic number (32?)

If I leave it to the default it seems too small, and I can't make it too big either. The default is 16, according to docs.

jessepeterson:
As far as I understand it makes it decode strictly gzip streams. Per the [zlib docs](http://www.zlib.net/manual.html#inflateInit2):

> windowBits can also be greater than 15 for optional gzip decoding. Add 32 to windowBits to enable zlib and gzip decoding with automatic header detection, or add 16 to decode only the gzip format (the zlib format will return a Z_DATA_ERROR). If a gzip stream is being decoded, strm->adler is a crc32 instead of an adler32.

I don't see it merged yet in main tree? Or are you saying you just merged it yourself locally for your testing? I, too, have around 50-60 recipes that seem to run just fine using this code.

timsutton:
Ah, ok. I had not gone so far as zlib's docs, only read the docs for the Python module. I'll add an explanation to the code and push the updated merge.

PR #187, opened 2015-06-16
Use GitHub's settings/tokens path rather than settings/applications.

eahrold:
I just noticed that when I crated an API token using `autopkg search -t ...` it listed the `AutoPkg CLI` token at https://github.com/settings/tokens but was not showing up at https://github.com/settings/applications. This simply updates the info string.


timsutton:
Thanks!

PR #194, opened 2015-07-16
Copier processor: also copy mode bits from file

jessepeterson:
Copy mode bits for a file that is copied with the Copier process. Note the `shutil.copy` copies the mode, as does `shutil.copytree`. To consider changing the code block to be simpler and just use `copy()` for anything not a directory.

### DIFF
--- a/Code/autopkglib/Copier.py
+++ b/Code/autopkglib/Copier.py
@@ -73,6 +73,7 @@ class Copier(DmgMounter):
                 shutil.copytree(source_item, dest_item, symlinks=True)
             elif not os.path.isdir(dest_item):
                 shutil.copyfile(source_item, dest_item)
+                shutil.copymode(source_item, dest_item)
             else:
                 shutil.copy(source_item, dest_item)
             self.output("Copied %s to %s" % (source_item, dest_item))


timsutton:
Interesting that this has only come up now, although I guess in many cases people normally set specific modes on files using e.g. the PkgCreator parameters. What was the specific case that led you to add this?

jessepeterson:
I'm overwriting a broken postinstall script from a package with a modified one (thanks, Extensis). So it needs the execute bits.

jessepeterson:
Also - can you think of any reason not to just fold the last elif and else cases into a single `shutil.copy2()`?

timsutton:
That seems reasonable. Not yet sure exactly what cases might be picked up by the last clause - symlinks perhaps?

timsutton:
I did worry about how this could affect other recipes, though, and the first one I tried, AdobeAIR.pkg, has issues with this change:

```
Copier
Copier: Copied /Users/tim/Library/AutoPkg/Cache/com.github.autopkg.pkg.AdobeAIR/AdobeAIR/Library/Frameworks/Adobe AIR.framework/Versions/Current/Adobe AIR Application Installer.app to /Users/tim/Library/AutoPkg/Cache/com.github.autopkg.pkg.AdobeAIR/AdobeAIR/Applications/Utilities/Adobe AIR Application Installer.app
Copier
Copier: Copied /Users/tim/Library/AutoPkg/Cache/com.github.autopkg.pkg.AdobeAIR/AdobeAIR/Library/Frameworks/Adobe AIR.framework/Resources/airappinstaller to /Users/tim/Library/AutoPkg/Cache/com.github.autopkg.pkg.AdobeAIR/AdobeAIR/Applications/Adobe/Flash Player/AddIns/airappinstaller/airappinstaller
Copier
Can't copy /Users/tim/Library/AutoPkg/Cache/com.github.autopkg.pkg.AdobeAIR/AdobeAIR/Library/Frameworks/Adobe AIR.framework/Resources/airappinstaller.rsrc to /Users/tim/Library/AutoPkg/Cache/com.github.autopkg.pkg.AdobeAIR/AdobeAIR/Applications/Adobe/Flash Player/AddIns/airappinstaller/airappinstaller/..namedfork/rsrc: [Errno 1] Operation not permitted: '/Users/tim/Library/AutoPkg/Cache/com.github.autopkg.pkg.AdobeAIR/AdobeAIR/Applications/Adobe/Flash Player/AddIns/airappinstaller/airappinstaller/..namedfork/rsrc'
Failed.
```

Haven't yet investigated whether this is something specifically to do the permissions of the resource forks of _this_ package or whether it would be a more general issue (either with resource forks or other weird files).

jessepeterson:
Yikes. Resources forks in 2015? Leave it to Adobe. :)

Okay, then I can update the PR to include a `copy_mode` key that can be turned on to enable this feature. That sound reasonable?

jessepeterson:
Alternatively we can just try/catch the copymode and silently fail.

timsutton:
I definitely wouldn't want to silently fail, but I'd also like to understand why the recipe above was failing on those resource forks.

In general I could see that copying out package resources that have restrictive modes _in the archive_, once sitting in the CACHE_DIR with those modes they may blow up when you try and manipulate them, whereas before when modes weren't being preserved we were sidestepping that issue.

Having an optional input variable seems like one solution - I'd be interested in seeing if there are other more complex recipes that involve Copier (maybe some things @nmcspadden or others have worked on that involve doing "surgery" on weird packages and repackaging) that would behave differently with modes being preserved.

jessepeterson:
I think the general problem is that resource forks can't have mode bits set on them:

```bash
% touch /tmp/foo
% echo bar > /tmp/foo/..namedfork/rsrc
% chmod 755 /tmp/foo/..namedfork/rsrc
chmod: Unable to change file mode on /tmp/foo/..namedfork/rsrc: Operation not permitted
```

So is that to say you'd favor a try/catch approach to copy modes by default but would want to see if anything breaks if that is changed (vs. an optional input var)?

jessepeterson:
Hey, @timsutton, did you have any thoughts for a direction to take? I'm happy to ammend the PR to include an optional switch for the copymodes, or add a try/catch like we were discussing. Thanks!

jessepeterson:
Couple more thoughts:

By default `cp` requires an extra switch to copy modes (`-p`). Finder, on the other hand, copies everything by default. Not sure which model should be followed.

Separately, If we were to copy modes by default, we can just exclude anything that has a `/..namedfork/` in the path.

novaksam:
I'm not sure if it's relevant, but I wrote a processor that copies over Icon data folder folders.

I'm using xattr to copy and write the data.

https://github.com/autopkg/novaksam-recipes/blob/master/Recipes%20-%20pkg/ResourceForkCopier.py

PR #195, opened 2015-07-16
Improve search results during `recipe run` by checking for exact match

eahrold:
When testing out the new interactive search feature during `autopkg run [recipe]`, I wasn't really getting any direct hits (i.e `len(repo_names) == 1` ) to take advantage of the `Add recipe repo 'xxx'? [y/n]:` feature. Usually it resulted in 5 or more results.

Here's a crack at improving the results by checking for exact matches based on either the file name or the recipe identifier. 

It adds two additional steps:
1. Iterate over the search results and compares the result item's "name" key to the "name" var used in the initial query. If an exact match is found it breaks and moves directly to the add repo prompt.
2. If an exact match is not found there, it then takes the top 5 results from the initial API search, and using the `html_url` key performs follow-up requests to `raw.githubusercontent.com`. The appeal of this technique is that it doesn't require calls to the API and consequently won't charge against the rate limit.

@gregneagle, you may have already gone down this road and decided any additional network calls were  too expensive, but I though I would put it out there for consideration.

As always any suggestions on how this could be improved are appreciated. 

Thanks

### DIFF
--- a/Code/autopkg
+++ b/Code/autopkg
@@ -247,8 +247,7 @@ def locate_recipe(name, override_dirs, recipe_dirs,
             % (indef_article, name))
         if answer.lower().startswith('y'):
             results_limit = 100
-            query = 'q="%s"+extension:recipe+user:autopkg+in:file,path' % name
-
+            query = 'q=%s+extension:recipe+user:autopkg+in:file,path' % name
             query += "&per_page=%s" % results_limit
 
             results = do_gh_code_search(query, use_token=False)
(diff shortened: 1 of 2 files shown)


gregneagle:
@eahrold: Can you provide some sample runs so I can compare behaviors?

IE: autopkg run foo triggering a search for foo, so I can see actual examples of the behavior.

Unseen/untested, I worry about hiding information from the user, so I'd like to see a demo. My initial tests haven't turned up anything "interesting"

gregneagle:
This test resulted in less than ideal behavior:
```
bash-3.2$ ./autopkg run VMwareFusion.jss
Didn't find a recipe for VMwareFusion.jss.
Search GitHub AutoPkg repos for a VMwareFusion.jss recipe? [y/n]: y
Nothing found.

Nothing downloaded, packaged or imported.
bash-3.2$ autopkg search VMwareFusion.jss

Name                         Repo                     Path                                    
----                         ----                     ----                                    
VMwareFusion.jss.recipe      golbiga-recipes          JSS/VMwareFusion.jss.recipe             
VMwareFusion.jss.recipe      justinrummel-recipes     VMware Fusion/VMwareFusion.jss.recipe   
VMware Fusion.jss.recipe     novaksam-recipes         VMWare Fusion/VMware Fusion.jss.recipe  

To add a new recipe repo, use 'autopkg repo-add <repo name>'
bash-3.2$ 
```

I think `autopkg run VMwareFusion.jss`, when it offered to search, should have at the least returned the same list that the `search` verb does.

eahrold:
Hmm, I ended up quoting the name parameter because when doing calls like "GitHub.jss" it was blowing out the results, but looks like that's also affecting the VMWareFusion.jss search too...

changing  `query = 'q="%s"+extension:recipe+user:autopkg+in:file,path' % name`
back to `query = 'q=%s+extension:recipe+user:autopkg+in:file,path' % name`

Gets the results list back to showing you're seeing back and also prompts to add
```
autopkg run VMwareFusion.jss
Didn't find a recipe for VMwareFusion.jss.
Maybe you meant VMwareTools.jss?
Search GitHub AutoPkg repos for a VMwareFusion.jss recipe? [y/n]: y

Name                         Repo                     Path                                    
----                         ----                     ----                                    
VMwareFusion.jss.recipe      golbiga-recipes          JSS/VMwareFusion.jss.recipe             
VMwareFusion.jss.recipe      justinrummel-recipes     VMware Fusion/VMwareFusion.jss.recipe   
VMware Fusion.jss.recipe     novaksam-recipes         VMWare Fusion/VMware Fusion.jss.recipe  

Add recipe repo 'golbiga-recipes'? [y/n]: 
```

eahrold:
But I see the trouble with this logic in that it's assuming the first exact match of a file name is the correct one.

gregneagle:
OK, but it's not clear why it decides to offer 'golbiga-recipes' and not 'justinrummel-recipes', both of which seem to contain a recipe with the exact name I'm looking for.

eahrold:
JINX

gregneagle:
It's hard to to this "correctly" which is why I didn't even try unless there was a single obvious match.

eahrold:
You're right. I added in the file name matching last, and now I'm thinking that routine should get cut, and simply do the check for identifiers. 

However it seems like the identifier check was only improving results when running into a recipe missing it's parent.

I'll go back to the drawing board and clean some of this up. Thanks.

eahrold:
So thinking about the VMwareFusion.jss delima a bit more. The search results do have a 'score' key based on what ever analytics GitHub has in play. Maybe it's more about informing.

So the file name match would stay in place, but show a message like

'Multiple recipes with matching names were found. The one with the highest "score" is included with the repo [some repo]' 

'Do you want to add this repo?'

gregneagle:
I'm not sure we should trust GitHub's analytics or the 'score' unless we understand better what it actually _means_.

eahrold:
Fair enough, and it's wonderfully vague...

>Ranking search results
>
>Unless another sort option is provided as a query parameter, results are sorted by best match, as indicated by the score field for each item returned. This is a computed value representing the relevance of an item relative to the other items in the result set. Multiple factors are combined to boost the most relevant item to the top of the result list.

eahrold:
@gregneagle, I updated this, restoring the previous query string (and functionality), and removing the filename match routine. 

Here's an example that demonstrates where it becomes useful

For demo purposes add the jss recipes and remove the 'recipes' repo (since it contains the parents) 

```
autopkg repo-add https://github.com/autopkg/jss-recipes.git
autopkg repo-delete recipes
```

Then run a recipe with a missing parent,

```
autopkg run Adium.jss

Didn't find a recipe for com.github.autopkg.pkg.Adium.
Search GitHub AutoPkg repos for a com.github.autopkg.pkg.Adium recipe? [y/n]: y

Name                   Repo            Path                                    
----                   ----            ----                                    
Adium.jss.recipe       jss-recipes     Adium/Adium.jss.recipe                  
Adium.munki.recipe     recipes         Adium/Adium.munki.recipe                
Adium.pkg.recipe       recipes         Adium/Adium.pkg.recipe                  

A recipe was found with an identifier matching 'com.github.autopkg.pkg.Adium' in the recipe repo 'recipes'

Add recipe repo 'recipes'? [y/n]: 
```

And while this wouldn't improve the results when running with the recipe name
```
autopkg run Adium.pkg
```

It would find an exact match if in some bizarre situation the recipe was run by identifier
```
autopkg run com.github.autopkg.pkg.Adium
```

I think this inches closer, let me know if you have other suggestion or concerns.

gregneagle:
We're going to release 0.5.0 today without these changes, but we'll continue to look at this. I can see these changes being useful and we may merge them in a follow-up release. We don't plan to wait six-seven months until the next release :-)

eahrold:
Great, 
I knew the new release was close and didn't expect this to make it in, just wanted to open a discussion.

eahrold:
💀

PR #204, opened 2015-07-26
Added recipe name to Munki import summary results

nmcspadden:
Adds the name of the recipe to the summary result output for Munki imports.  This also has the effect of adding it to the reporting plist, so that Failures, URL Downloads, and Munkiimports all report the recipe name.

### DIFF
--- a/Code/autopkg
+++ b/Code/autopkg
@@ -1502,12 +1502,14 @@ def run_recipes(argv):
                     result = item["Output"][key]
                     summary_text = result.get('summary_text', '')
                     data = result.get('data')
+                    data['recipe_name'] = recipe_path
                     if key not in summary_results:
                         summary_results[key] = {}
                         summary_results[key]['summary_text'] = summary_text
                         if type(data).__name__ in ['dict', '__NSCFDictionary']:
                             summary_results[key]['header'] = (
                                 result.get('report_fields') or data.keys())
+                            summary_results[key]['header'].append('recipe_name')
                         summary_results[key]['data_rows'] = []
                     summary_results[key]['data_rows'].append(data)
 


timsutton:
I'm seeing duplicated listings in URLDownloader summary results:

```
./autopkg run --report-plist out XLD.munki VLC.munki -v

...

The following new items were imported into Munki:
    Name  Version   Catalogs  Pkginfo Path                    Pkg Repo Path                 Recipe Name
    ----  -------   --------  ------------                    -------------                 -----------
    XLD   20141129  testing   apps/XLD/XLD-20141129__1.plist  apps/XLD/XLD-20141129__1.dmg  XLD.munki
    VLC   2.2.1     testing   apps/VLC/VLC-2.2.1.plist        apps/VLC/VLC-2.2.1.dmg        VLC.munki

    Recipe Name  Download Path                                                                    Recipe Name
    -----------  -------------                                                                    -----------
    XLD.munki    /Users/tim/Library/AutoPkg/Cache/com.github.autopkg.munki.XLD/downloads/XLD.dmg  XLD.munki
    VLC.munki    /Users/tim/Library/AutoPkg/Cache/com.github.autopkg.munki.VLC/downloads/VLC.dmg  VLC.munki
```

The plist output only lists the Recipe Name twice in the `header` key, not in `data_rows`:
```
		<key>url_downloader_summary_result</key>
		<dict>
			<key>data_rows</key>
			<array>
				<dict>
					<key>download_path</key>
					<string>/Users/tim/Library/AutoPkg/Cache/com.github.autopkg.munki.XLD/downloads/XLD.dmg</string>
					<key>recipe_name</key>
					<string>/Users/tim/Library/AutoPkg/RecipeRepos/com.github.autopkg.recipes/XLD/XLD.munki.recipe</string>
				</dict>
			</array>
			<key>header</key>
			<array>
				<string>recipe_name</string>
				<string>download_path</string>
				<string>recipe_name</string>
			</array>
			<key>summary_text</key>
			<string>The following new items were downloaded:</string>
		</dict>
```

Since 'name' can be a full path, output can get very wide, but I think very few people usually run recipes by paths.

I think there's something to the idea of being able to put more detailed output that's reserved _exclusively_ for plist output, for recording paths and identifiers (similar to how diskutil/hdiutil gives more info in plist output than regular output), but since the current report plist output structure is more directly connected to the summary display output that might be awkward to decouple.

jessepeterson:
👍  Perhaps a mechanism to do this is with a simple underscore that precedes the output row name. Those can be omitted from the output data.

nmcspadden:
I don't really need this anymore, and I'm going to rethink if it's even useful.

PR #208, opened 2015-08-03
SparkleUpdateInfoProvider: Raise ProcessorError if channel items list empty

jps3:
* `get_feed_data()` did not check that any `channel/item` items were
actually found/returned, which would lead to an `IndexError` in `main()`.
* Discovered when running for  'com.github.jps3.download.fseventer’
which is returning `200 OK` for a `text/html` document (contents of
which simply state that site is temporarily offline). Hence the Processor
would chug along without a care in the world.

### DIFF
--- a/Code/autopkglib/SparkleUpdateInfoProvider.py
+++ b/Code/autopkglib/SparkleUpdateInfoProvider.py
@@ -144,10 +144,8 @@ class SparkleUpdateInfoProvider(Processor):
         except:
             raise ProcessorError("Error parsing XML from appcast feed.")
 
-        try:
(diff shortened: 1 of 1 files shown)


timsutton:
0.5.1 doesn't have the issue because we now wrap Processors in an attempt to catch unhandled exceptions, but it would still be good to have a sanity check like this, even though it probably affects very few sparkle feeds.

jps3:
Ok, I think I did this pull request update correctly. (I had started to make a hash of things by merging in upstream updates and this change, but it looks like I reset, updated and pushed changes correctly.)

timsutton:
Perfect. Thanks!

PR #209, opened 2015-08-04
Fix Changelog links

clburlison:
Now all links actually work.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
-### [0.5.1][unreleased] (Unreleased)
+### [0.5.1](https://github.com/autopkg/autopkg/compare/v0.5.0...HEAD) (Unreleased)
 
-### [0.5.0][] (July 17, 2015)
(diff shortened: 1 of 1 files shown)


homebysix:
Oh interesting. GitHub sees `[something] (something)` and assumes the space between is a mistake. My markdown preview package doesn't do that. Good catch.

Looks like you lost the space before the date on version 0.4.1, though.

timsutton:
Have you already tested simply:

`[0.5.0](https://github.com/autopkg/autopkg/compare/v0.4.2...v0.5.0)  (July 17, 2015)`

..to avoid having to maintain the list of links at the bottom?

clburlison:
@timsutton PR is updated with that markdown syntax. Hopefully that makes it easier to maintain.

timsutton:
Thanks, I cherry-picked these commits to avoid a merge. Going back to look at the release script turned up something else I needed to update anyway.

PR #212, opened 2015-08-11
MunkiImporter empty extension fix

hewei100:
Empty pkginfo extension is supported by munkiimport but autopkg doesn't handle it well. This fix took munkiimport's original codes as reference and has been working for me for a year.


gregneagle:
Looks fine. Thanks!

PR #215, opened 2015-08-14
Removed trailing colon from URL in error messages.

homebysix:
This will allow tools that email AutoPkg error output (like AutoPkgr) to produce clickable URLs. Current behavior causes a trailing colon to render the URL unusable.

screen shot 2015-08-14 at 9 20 38 am

This change will remove the trailing colon from the URL, and surround the error message with parentheses to call it out.

### DIFF
--- a/Code/autopkglib/URLDownloader.py
+++ b/Code/autopkglib/URLDownloader.py
@@ -242,7 +242,7 @@ class URLDownloader(Processor):
 
         except BaseException as err:
             raise ProcessorError(
-                "Couldn't download %s: %s" % (self.env["url"], err))
+                "Couldn't download %s (%s)" % (self.env["url"], err))
         finally:
             if url_handle is not None:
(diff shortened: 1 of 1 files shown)


gregneagle:
I'm not sure it's really needed that URLs in error messages be clickable (after all, one of the reasons for errors is bad/invalid URLs), but this seems OK. The intent was that the error message read reasonably like an English sentence without being too verbose and this seems to preserve that.

gregneagle:
For completeness, will you make similar changes to the other included processors that might have similar errors?

homebysix:
@gregneagle I made a pass through the other processors and didn't see any similar trailing colons used after URLs.

gregneagle:
I'm thinking stuff like CURLDownloader and URLTextSearcher and CURLTextSearcher, but if they need tweaking that can happen down the line I guess.

PR #216, opened 2015-08-19
MunkiImporter: Try to match bundle types as well as application types…

jessepeterson:
Currently MunkiImporter only looks for applications when trying to match the `installs` keys with previous versions. This PR allows it to also check `bundle` types. Came up in autopkg/jessepeterson-recipes#17


timsutton:
Thanks!

PR #220, opened 2015-08-26
PkgCreator path joining

nmcspadden:
Initial fix for allowing PkgCreator to handle absolute paths in a chown dict.  Currently if an absolute path is provided, autopkgserver will attempt to change owner, group, and mode on that absolute path.  It should always only ever touch things within the pkgroot.

### DIFF
--- a/Code/autopkgserver/packager.py
+++ b/Code/autopkgserver/packager.py
@@ -358,7 +358,11 @@ class Packager(object):
                      str(entry.user),
                      str(entry.group)))
 
-            chownpath = os.path.join(self.tmp_pkgroot, entry.path)
+            if entry.path.startswith('/'):
+                chownpath = os.path.join(self.tmp_pkgroot,
+                    entry.path.lstrip('/'))
+            else:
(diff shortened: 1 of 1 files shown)


Code/autopkgserver/packager.py
```diff
@@ -358,7 +358,11 @@ def verify_relative_valid_path(root, path):
                      str(entry.user),
                      str(entry.group)))
 
-            chownpath = os.path.join(self.tmp_pkgroot, entry.path)
+            if entry.path.startswith('/'):
+                chownpath = os.path.join(self.tmp_pkgroot,
+                    entry.path.lstrip('/'))
+            else:
+                chownpath = os.path.join(self.tmp_pkgroot, entry.path)
```
gregneagle: Why not just the simpler:

`chownpath = os.path.join(self.tmp_pkgroot, entry.path.lstrip('/'))`

to replace line 361 and not have all the conditional stuff?  Are there any other os.path.join calls that might have similar issues?

PR #229, opened 2015-09-29
Move traceback capturing to autopkglib/__init__.py. Update output.

sheagcraig:
- Get the traceback from autopkglib/**init**.py.
  - The traceback captured by `traceback.format_exc`, in autopkg is
    not the traceback we're interested in seeing.
- Add printing of the traceback if the verbose level is greater than 1.
- Reformat failure printing to be slightly easier to read (break into
  multiple lines of data).

### DIFF
--- a/Code/autopkg
+++ b/Code/autopkg
@@ -40,7 +40,6 @@ import re
 import shutil
 import subprocess
 import time
-import traceback
 import autopkglib.github
 
 from urlparse import urlparse
@@ -1465,7 +1464,7 @@ def run_recipes(argv):
             log_err("Failed.")
             failure["recipe"] = recipe_path
             failure["message"] = unicode(err)
-            failure["traceback"] = traceback.format_exc()
+            failure["traceback"] = err.traceback
             failures.append(failure)
             autopackager.results.append({'RecipeError': unicode(err).rstrip()})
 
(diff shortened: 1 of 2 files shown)


sheagcraig:
I noticed that the traceback wasn't getting included at all in failure reports. Once I added it in, I realized that it was grabbing the wrong traceback, which wasn't super helpful. While exception handling is much hardier now, it did make it harder to troubleshoot issues with users (on Slack!).

There are a couple of issues:
- Rather than adding the exception with `__setattr__`, it would be better to add an exception property to the __init__ of the AutoPackagerError class. That seems like the right thing to do, but I wanted to see what your thoughts were in general, first.
- Do you feel like a verbosity level of 2 is appropriate for including the traceback?

sheagcraig:
For whatever reason I thought that using `__setattr__` would be useful for indicating that the attribute "traceback" was being set in a non-standard way (i.e., it didn't actually exist).

But it can just as easily be added by assigning to it. Perhaps that's better, with a comment? Or just add it with a default value of None to the `__init__` for the exception class.

sheagcraig:
@timsutton Any word on this? I've rather lost track of it since it's been about 3/4 of a year. Let me know how I can help push this through.

PR #236, opened 2015-11-10
Merge versioning processors

sheagcraig:
While working on recipe robot I was reminded that I don't see the need for both a Versioner and an AppDmgVersioner (aside from backwards compatibility with recipes using one or the other).

I combined the features of the two. I updated Versioner since that seems like the more appropriate target for the final processor.

So Versioner gains the output variables app_name and bundleid.

Also, AppDmgVersioner can search within an image for the first app it finds. I added this ability to Versioner, but also added the ability to search for other bundle types. The only ones I could come up with were .plugin and .qlgenerator, although I'm sure there are others that could be added to the list. The final code prioritizes apps, so you don't have to worry about breaking older recipes that may include both at the base of an image.

Finally, the code has been refactored to put exception handling and disk image mounting/unmounting into smaller try/except/finally blocks. AppDmgVersioner used a too-broad `except BaseException` to protect against failed dictionary lookups, which has been corrected.

The code passes pylint.


timsutton:
Thanks! I think as long as there is a "versioning-specific" processor there may as well be only one. There should be quite a few recipes using AppDmgVersioner already, so a good way to test this would be to do sub AppDmgVersioner for your updated Versioner on a corpus of recipes and verify that we get the same results.

If things look good, we could also update docs for Versioner that it's deprecated and eventually nudge people to migrate.

sheagcraig:
Do you want me to use existing recipes to test, or can I just write up a generic test to demonstrate their swappability?

You mean update docs for AppDmgVersioner that say it's deprecated, right?

sheagcraig:
Unless you think it's a bad idea, I'm going to add code to use the AppDmgVersioner input variable names and map them to the Versioner names. That way somebody could just change the processor name to update.

sheagcraig:
Actually, I take that back. There's no good way to ensure Versioner has an input_plist_path while allowing the dmg_path from AppDmgVersioner without updating the autopkglib.__init__ code that ensures required input variables are present.

Unless you've got any ideas.

sheagcraig:
I don't see any tests sitting around anywhere, so I started a Tests folder. You can run them with `python test_versioner.py`.

I normally use nose as a test runner, but to avoid introducing dependencies I just used unittest. This means some funny business about importing autopkglib that requires the tests to be run from the test directory. I'm sure there's a better way to do this; and arguably, for developers, I don't know if it would be bad to just use nose.

Finally, because I'm strapped for development time, I just edited an Info.plist and made a few pretend app, plugin, and disk images containing them for testing. This would be better handled by using mock to patch FoundationPlist.readPlist and the dmg mounting code, but that's both another dependency, and not  an efficient use of my time unless we get serious about testing.

Anyway, the first test specifically tests for whether Versioner will do the same thing that AppDmgVersioner does. The rest are just testing ~~all of~~ most of the permutations of input.

sheagcraig:
I neglected to put in tests for the missing information-they're in now. Specifically, AppDmgVersioner would report UNKNOWN_VERSION and UNKNOWN_BUNDLEID if either was missing. Also, it now raises a ProcessorError if it can't find a bundle in the first place.

As it stands, this is possible if you give it a path to a plist file that is not named Info.plist, or not contained within a bundle of one of the known types. If this needs to be more flexible, I can open it up more.

sheagcraig:
For what it's worth, for the tests I have "actual" files; if it's okay to use mock.patch (non std-lib) I can remove all of that stuff and just patch FoundationPlist.readPlist and DmgMounter.mount.

timsutton:
Thanks for your work on this. A project like AutoPkg would really benefit from being able to test processor behaviour in an automated way, especially considering its use is based heavily upon user input and servers we don't control (the latter of which are also harder to mock up I'd think).

I don't really have an opinion on whether or not to use mock. I'm not really opposed to having external package requirements for things like unit tests since the common packages for that aren't obscure and they aren't needed except for running the tests (which a bot could do). I also don't think it's a big deal to leave the fake resources in the repo currently, as they compress down to nothing and could be removed later.

As for how to handle "dynamically-assigned" input variables, I don't have any ideas at the moment. I know this came up previously for things like PlistReader and URLTextSearcher.

sheagcraig:
Any updates on whether this should be merged in or not?

PR #241, opened 2015-11-24
Add support for a GitHub or BitBucket repo's SSH URL 

eahrold:
This treats strings prefixed with `git@` as a valid clone URL.

Additionally I extracted the checks that use urlparse() into a new `valid_repo_url()` function.
The one difference I noticed was that in the `expand_repo_url()` func the scheme was checked, where as in `get_repo_info()` netloc was checked. 

They seem to both be used to validate a string as a URL, but there may be an aspect you've thought through that I'm missing. I can roll that back if need be.

### DIFF
--- a/Code/autopkg
+++ b/Code/autopkg
@@ -162,6 +162,9 @@ def valid_override_file(filename):
     override_plist = recipe_plist_from_file(filename)
     return valid_override_plist(override_plist)
 
+def valid_repo_url(url):
+    '''Returns True if the url is a git clone url, otherwise False'''
+    return True if url.startswith('git@') or urlparse(url).netloc else False
 
 def find_recipe_by_name(name, search_dirs):
     '''Search search_dirs for a recipe by file/directory naming rules'''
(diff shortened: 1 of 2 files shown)


gregneagle:
Aren't the SSH URLs more properly written: `ssh://git@github.com/path/to/repo`?

If we're going to support the alternative SSH URL syntax described in `git clone --help`, why limit it to URLs that start with `git@` ?  That's certainly not a requirement of the syntax.

eahrold:
`ssh://git@github.com...` would work anyway since that breaks up correctly with urlparse().
This just additionally adds support for `git@...` as a url, as advertised by GitHub/BitBucket.

ssh-url

Taking a close look, however, a reverse domain for the local path isn't constructed correctly as a reverse domain in either case due to the `@git` I'll tackle that next.

gregneagle:
So we either tell people to use the ssh:// syntax, or we implement "correct" support for the alternative syntax, in which "git@" is not special.

eahrold:
So a generic _user@_ ?

gregneagle:
Quoting from the output of `git clone --help`:

```
An alternative scp-like syntax may also be used with the ssh protocol:

       o   [user@]host.xz:path/to/repo.git/

       This syntax is only recognized if there are no slashes before the first colon. This helps differentiate a local path that contains a colon. For example the local
       path foo:bar could be specified as an absolute path or ./foo:bar to avoid being misinterpreted as an ssh url.
```

Notice that the `user@` part is _optional_!

eahrold:
@gregneagle,  This PR should now support full implementation of all clone syntax listed in `git clone --help`. Let me know if you have any other concerns. Thanks.

eahrold:
💀

PR #245, opened 2016-01-05
Switch to fetch and reset for repo-update

keeleysam:
This changes the `autopkg repo-update` command to use `git fetch` and `git reset` instead of `git pull`. It slightly changes the behavior, but will make it seamless if a repo is force updated.  It is also the behavior that most would expect from `git pull` if they have never had conflicts.

### DIFF
--- a/Code/autopkg
+++ b/Code/autopkg
@@ -771,9 +771,10 @@ def repo_update(argv):
                 repo_dirs.append(repo_path)
 
     for repo_dir in repo_dirs:
-        log("Attempting git pull for %s..." % repo_dir)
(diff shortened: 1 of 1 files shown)


gregneagle:
If an autopkg user makes modifications to a local copy of a recipe and does `autopkg repo-update`, git will warn if there are local changes that haven't been committed. This serves as a reminder so people don't lose important changes.

Does your proposed change alter this behavior, and if so, how?

timsutton:
Maybe we can make `repo-update`'s behaviour more intentional and just abort it with a warning if that repo's working copy is not in a clean state?

PR #250, opened 2016-02-19
just return bools of test

rmanly:
In two places there was if/then or assignment when all that was needed was a return of bool.


timsutton:
Thanks!

PR #254, opened 2016-03-03
Added functionality to determine the remote filename of the URL …

n8felton:
… being downloaded.

Not all URLs end with an appropriate filename. This allows for determining
the effective remote filename of a given URL without needing to use the
'filename' input variable. Additionally, the AutoPkg cache will now keep
multiple updates of a download rather than overwrite previous versions
when using the 'filename' input variable.

### DIFF
--- a/Code/autopkglib/URLDownloader.py
+++ b/Code/autopkglib/URLDownloader.py
@@ -22,6 +22,8 @@ import time
 import xattr
 import tempfile
 
+from urllib2 import unquote
+
 from autopkglib import Processor, ProcessorError
 try:
     from autopkglib import BUNDLE_ID
@@ -113,6 +115,36 @@ class URLDownloader(Processor):
         },
     }
 
+    def _curl_filename(self, curl_args, curl_path=None):
+        if curl_path is None:
+            curl_path = [self.env['CURL_PATH']]
+        curl_cmd = curl_path + curl_args
+        self.output(' '.join(curl_cmd), verbose_level=2)
+        proc = subprocess.Popen(curl_cmd, shell=False, bufsize=1,
+                                stdin=subprocess.PIPE,
+                                stdout=subprocess.PIPE,
+                                stderr=subprocess.PIPE)
+
+        file_url = proc.stdout.readline()
+        filename = file_url.rpartition("/")[2]
+        return file_url, filename
+
+    def _remote_filename(self, url):
+        curl_args = ['--silent',
+                     '--location',
+                     '--head',
+                     '--write-out', '%{url_effective}',
+                     '--url', url,
+                     '--output', '/dev/null']
+
+        (file_url, filename) = self._curl_filename(curl_args)
+
+        # Decode any special characters in the filename, like %20 to a space.
+        filename = unquote(filename)
+        self.output("Found filename '%s' at '%s'" % (filename, file_url),
+                    verbose_level=2)
+        return filename
+
     def main(self):
         # clear any pre-exising summary result
         if 'url_downloader_summary_result' in self.env:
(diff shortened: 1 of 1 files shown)


arubdesu:
Clever, good work! :+1:

homebysix:
Neat! Mind if I steal it for [Recipe Robot](https://github.com/homebysix/recipe-robot)?

n8felton:
@arubdesu Thanks!
@homebysix Sure thing!

timsutton:
I get the idea, although I'm not _totally_ convinced it's a solution for all servers that redirect through CDNs, etc. I also find there are some servers that also give different (or error) responses with `HEAD`. Do you have some examples of recipes for whose behaviour this would fix what you consider problematic? (I say "consider problematic" because it doesn't bother me that I may wind up with a download called lhasd8fo23rahsdi8fha298fh2fnasdfia_fhosdfh92 in my cache, if I know I can trust its contents to follow a pattern - I don't see that random name propagate any further in the processing flow)

About it resulting in multiple updates, this is one behaviour that would change as well, for better or for worse. In general, most recipes overwrite their own cached data with every run, with a few exceptions (like Office 2016 updates recipes..)

gregneagle:
Intended filenames can also by indicated in the Content-Disposition header if it exists. I think that would be more "canonical" than relying on the tail of the URL, even post all redirects.

n8felton:
> Do you have some examples of recipes for whose behaviour this would fix what you consider problematic?

One of the reasons I began looking at this is that I wish to create recipes for vendor provided printer drivers. Having to deal with mostly Xerox where I work, I started there. After following the support page rabbit hole, you end up at `http://www.support.xerox.com/support/workcentre-7220-7225/file-redirect/enus.html?contentId=131985`

If you throw this into the URLDownloader processor, without setting the `filename` input variable, you end up with a filename of `enus.html?contentId=131985` which results in the error `[...]/enus.html?contentId=131985 is not a valid installer item!` when trying to use the munki recipe. This PR will help this recipe, and other recipes with similar download URLs, translate the `pathname` output variable from URLDownloader into a usable filename.

Yes, I could set the `filename` input variable, but I always felt like that was more of a workaround for these types of URLs. At the end of it all, the `filename` input variable, if set, will still be used first. This helps with backward compatibility of recipes that already set the value. This code just adds a bit more logic behind attempting to grab a file name rather than just looking at the tail end of URL for future recipes, or for those who wish to override/remove the `filename` input variable and store the filenames as they are served from the servers.

> About it resulting in multiple updates, this is one behavior that would change as well, for better or for worse. In general, most recipes overwrite their own cached data with every run, with a few exceptions (like Office 2016 updates recipes..)

As mentioned above, an explicitly set `filename` input variable will override this code. For the recipes that already have this variable set, and for anyone that wishes to override recipes that do not have it set, this code should not interfere, continuing the current behavior of over writing files with the same filename in the AutoPkg cache.

Those of us that are digital pack rats, especially when storage is not an issue, storing old versions of installers is sometimes the more desired behavior, which requires the unique filenames that this code can help provide.

> Intended filenames can also by indicated in the Content-Disposition header if it exists. I think that would be more "canonical" than relying on the tail of the URL, even post all redirects.

I actually started down this path at first, but then came across some interesting discoveries. Mostly, not many people use Content-Disposition in their headers. Running through all the recipes in `autopkg/recipes`, I only found that Dropbox and the Office2016 recipes actually used the Content-Disposition header. Spot checking those recipes, the output of this code was an exact match of those recipes Content-Disposition value.

This is probably due to some of the reasons described in places like http://stackoverflow.com/questions/1012437/uses-of-content-disposition-in-an-http-response-header and http://www.gnucitizen.org/blog/content-disposition-hacking.

I did originally have code that would check if Content-Disposition was present first (`curl --location --head --silent --remote-name --remote-header-name --write-out %{filename_effective} --url "https://www.dropbox.com/download?plat=mac&full=1"`), and if not, fall back to the code that in in this PR. I felt that because of so few sources using the Content-Disposition header, this was just adding additional curl requests without any added benefit.

/end wall of text

n8felton:
Hey all, looking to bump this PR. Are there any additional questions I can help to answer?

n8felton:
Pending the work of #544 to be completed so that I can reevaluate this request and look at necessary code changes.

MichalMMac:
I would love this feature to be implemented somehow. I would also like to help wit this but six opened PRs is enough for the time being.

PR #257, opened 2016-03-12
Xip Archive expander

oucsaw:
Although having only seen one xip package in the wild, I'm not entirely
convinced of the value of this, however,  this commit contains an
XipArchiveExpander processor for use with .xip files, that will expand
them into a given directory.

See autopkg/jps3-recipes#22 (comment)
for more details.

### DIFF
--- a/Code/autopkglib/XipArchiveExpander.py
+++ b/Code/autopkglib/XipArchiveExpander.py
@@ -1,6 +1,6 @@
 #!/usr/bin/python
 #
-# Copyright 2016 Aaron Wilson
+# Copyright 2016 University of Oxford
 # Based on PkgExtractor.py, Copyright 2013 Greg Neagle
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


gregneagle:
https://github.com/autopkg/gregneagle-recipes/tree/master/SuspiciousPackageXip

These recipes handle that xip file without needing a new processor, but I cheated and modified some existing processors. At the very least we need to modify CodeSignatureVerifier to verify xip archives.

I don't feel strongly one way or the other about this processor; certainly if we start seeing more xip archives it will be useful and easier to work with than the techniques I used in my recipes.

timsutton:
I also thought that at least it may make more sense to add this functionality to the existing Unarchiver processor.

If the processor detects a xip by extension or is explicitly told it's a xip, it could pass in the desired arguments to `xar` as Greg did here: https://github.com/autopkg/gregneagle-recipes/blob/master/SuspiciousPackageXip/SuspiciousPackageXip.pkg.recipe#L25-L26

oucsaw:
I thought about adding this to the Unarchiver, but wasn't 100% convinced that this was the correct way to go - I don't have really strong feelings about this, mind.

Given that there is a work around, I will close this and if I get the chance look to add this to the Unarchiver.

PR #263, opened 2016-03-30
Add DMG mounter to FileFinder and provide DMG-relative file finding

jessepeterson:


jessepeterson:
Just a note that this PR is backwards compatible — does not change behavior for existing `FileFinder` users. I'm happy to discuss it, the implementation, or anything like that.

gregneagle:
LGTM

PR #264, opened 2016-04-03
Add kext codesignatureVerification

arubdesu:
Currently the CodeSigVerifier processor only allows app bundles to have the code signature checked. This expands file types to allow signed kexts to be checked as well.

### DIFF
--- a/Code/autopkglib/CodeSignatureVerifier.py
+++ b/Code/autopkglib/CodeSignatureVerifier.py
@@ -254,7 +254,7 @@ class CodeSignatureVerifier(DmgMounter):
 
             # Currently we support only .app, .pkg, .mpkg or .xip types
             file_extension = os.path.splitext(matched_input_path)[1]
-            if file_extension == ".app":
+            if file_extension in [".kext", ".app"]:
                 self.process_app_bundle(matched_input_path)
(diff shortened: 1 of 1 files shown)


arubdesu:
Paging @hjuutilainen

gregneagle:
Where are you finding a "naked" kext? They should be inside signed pkgs....

arubdesu:
Santa's pkg isn't signed at present, so I'd need to check their .app bundle and .kext individually. I'm [commenting out](https://github.com/autopkg/arubdesu-recipes/blob/dfbcd7df4c6720b5437537434b07189c700801ae/santa/santa.download.recipe#L78-L88) the relevant part of the recipe for now, and have [pinged](https://github.com/google/santa/issues/40) those folks to see if they could release signed pkgs in the future.

hjuutilainen:
I'm fine with this change. I'd even support the decision to only check file extensions for passing to pkgutil and feed everything else to codesign. At some point I'd like to separate these two kind of signature checks to their own processors but that's a discussion for a whole different PR.

arubdesu:
This is resolved in [my recipe](https://github.com/autopkg/arubdesu-recipes/blob/master/santa/santa.download.recipe), we can close this and say 'whenever Hannes is inspired to update the processor he'll do so'.

PR #266, opened 2016-04-12
Improved PkgCopier default pkg_path

magnusviri:
Currently if a glob character is used (like *.pkg) and a default pkg_path isn't defined, then the pkg_path will be "*.pkg".

This change switches the default from source_pkg to matched_source_path.  This way if glob character are used the pkg_path will use the actual file name that was matched and not the glob pattern.

### DIFF
--- a/Code/autopkglib/PkgCopier.py
+++ b/Code/autopkglib/PkgCopier.py
@@ -87,7 +87,7 @@ class PkgCopier(Copier):
             # do the copy
             pkg_path = (self.env.get("pkg_path") or
                         os.path.join(self.env['RECIPE_CACHE_DIR'],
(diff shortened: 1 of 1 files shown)


timsutton:
Thanks, this looks like a reasonable change. I'm curious, which recipe were you using/writing that exposed this issue?

magnusviri:
As I write more and more recipe's and look at other people's I'm trying different ways of getting things done.  At first I was writing a recipe for drivers for the KeySpan USB driver.  I eventually got around it with a custom processor I wrote to get version numbers from paths, but there still might be another better way of doing it.  When I finish writing all of my recipes I might go back over all of them and see if I can write them all better.

PR #270, opened 2016-04-19
StopIfVersionUnchanged.py: check version to end recipe early

ftiff:
My workflow involves working with large zip files on remote shares. 
My goal with this processor is to end recipe execution as soon as I get the version to save resources.

elios had an interesting view on this Processor that we could use as a conversation starter:
> Just be aware that stopping can lead to unexpected behaviors for child recipes
> Example: I make a Munki recipe that has your recipe as a parent.
> I run it, and it imports into Munki.
> There is some issue and I delete the stuff it imported.
> If I run the Munki recipe again, expecting it to re-import the item, it won't.
> It will stop early since the download didn't change.

### DIFF
--- a/Code/autopkglib/StopIfVersionUnchanged.py
+++ b/Code/autopkglib/StopIfVersionUnchanged.py
@@ -54,6 +54,8 @@ class StopIfVersionUnchanged(Processor):
             with open(path, 'r') as file:
                 file_content = file.read()
                 return BundleInfo._make(re.search(p, file_content).group(1, 2, 3, 4))
+        except IOError:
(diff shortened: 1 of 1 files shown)


gregneagle:
I happen to agree with elios's comments. I wrote the original StopProcessingIf processor, and later came to realize it was a bad idea.

I think for now I'd prefer it if you implement it as a Shared Processor in your own RecipeRepo. There's no need for it to be included with the "core" AutoPkg processors in order for you (and others) to use it if you choose.

ftiff:
Thanks @gregneagle I understand your concern. I guess it's a good time for me to finally work on my own repo.

PR #277, opened 2016-05-02
Fixed SparkleUpdateInfoProvider ignoring appcast_request_headers

jps3:
* It would appear that the recent change from using urllib to curl
introduced a subtle bug wherein the SparkleUpdateInfoProvider processor
was effectively ignoring `appcast_request_headers`.
* The fix: within the `get_feed_data()` function, the line calling
`self.fetch_content()` was giving `appcast_query_pairs` as the second
argument (which `get_feed_data()` took as its `headers=` argument).
Since `appcast_query_pairs` are url encoded in the code block
immediately before this call, all that’s missing was to remove the
extraneous argument with a corrected `headers=` one.


jps3:
Bump. 

(This affects, at least, com.github.timsutton.munki.HopperDisassembler3 and com.github.keeleysam.recipes.Twocanoes.WinclonePro4.munki).

timsutton:
Thanks for catching this!

PR #284, opened 2016-06-20
URLDownloader.py: stop processor on bad status from curl

oucsaw:
Add the --fail option to curl to make 400 errors (in particular 404
errors) return a non-zero exit code

Return an error from the processor when curl exits with a non-zero
code (i.e. error)

(as mentioned as an option in pr #273 )

### DIFF
--- a/Code/autopkglib/URLDownloader.py
+++ b/Code/autopkglib/URLDownloader.py
@@ -240,7 +240,7 @@ class URLDownloader(Processor):
                     break
 
         retcode = proc.poll()
-        if retcode: # Non-zero exit coudse from curl => problem with download
+        if retcode: # Non-zero exit code from curl => problem with download
             curlerr = ''
             try:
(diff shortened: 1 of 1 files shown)


Code/autopkglib/URLDownloader.py
```diff
@@ -240,19 +240,15 @@ def main(self):
                     break
 
         retcode = proc.poll()
-        if retcode:
+        if retcode: # Non-zero exit coudse from curl => problem with download
```
gregneagle: What is the word "coudse" supposed to be?

erikng: "Code".

gregneagle: So @oucsaw -- please fix the typo...

oucsaw:
It should be 'code'; now fixed.

Apologies for not reading the comments ...

PR #285, opened 2016-06-21
Allow for Less Strict "legal" Package Names.

goobersmooch:
Updated the re_pkgname regex to allow for a + sign in the package name.
This change will allow your name to contain a + sign when creating a
app.pkg.recipe without throwing an invalid package name error.


gregneagle:
See the prior discussions about the "legal" package name characters.
I don't think we want to just add random characters at random intervals. I'd rather see a good discussion and then one big change making the re more liberal.

Issue #296, opened 2016-07-13
AutoPkg shouldn't exit 1 if a recipe fails


nmcspadden:
Currently, if AutoPkg runs a recipe with 1 success, 1 fail, it returns an exit code of 1.

This is problematic in a number of ways:

- Based on exit code, it's not possible to tell if AutoPkg ran all recipes or failed due to some exception or traceback.
- AutoPkg successfully handling a failed recipe and moving on to the next one is expected behavior, and therefore should exit 0 if this happens.
- The reporting options (such as --reportplist) are there to inform the admins if recipes fail during the run, and even if a successful report is generated that contains fails, the exit code of 1 still seems to indicate an overall execution fail, even though the program behaved exactly as intended.

Exit non-zero should only happen in a case where AutoPkg encountered an error it couldn't resolve (such as a recipe being malformed, permissions issues, etc.).  A recipe failing due to ProcessorError exceptions, which are then skipped, should not indicate a program fail.

An autopkg run execution that only contains one recipe and fails that recipe should still exit 0 if it generated a report successfully and informed the user of the fail.

timsutton:
I think I agree. I'm betting that its behaviour for exiting non-zero if any failures occur may have predated us having the reporting option.

timsutton:
I'm guessing that the exit code is actually the number of failures, from looking at this: https://github.com/autopkg/autopkg/blob/master/Code/autopkg#L1583

timsutton:
I'm in favour of changing this to be that a successful run _completion_, regardless of individual recipe results, causes `run_recipes()` to exit 0.

I imagine this could break someone's AutoPkg run wrapper script, however I think if a script were doing that it wouldn't be very helpful, since you'd probably want to know _which_ recipe(s) failed anyway.

nmcspadden:
I'm in favor of that too. Would simply removing that one line be sufficient? It doesn't seem to be used anywhere else outside that function anyway.

gregneagle:
I'm still not really understanding the issue. (or maybe _why_ it's an issue)

gregneagle:
I think overall I'd be happier if it was something like 0 meant no errors, 1 meant one or more recipes failed, and -1 meant some other autopkg problem occurred.

I think intentionally returning 0 even if a recipe failed seems odd.

nmcspadden:
In my specific use case, I am wrapping AutoPkg runs with another script that also does a lot of other organization-specific things.  

[...]

PR #297, opened 2016-07-14
Autopkg exit codes

nmcspadden:
`autopkg run` now returns 1 if any recipe failed, rather than returning the number of recipes that failed.

### DIFF
--- a/Code/autopkg
+++ b/Code/autopkg
@@ -1580,7 +1580,7 @@ def run_recipes(argv):
         write_plist_exit_on_fail(results_report, options.report_plist)
         log("\nReport plist saved to %s." % options.report_plist)
 
-    return error_count
+    return error_count > 0
 
 
 def printplistitem(label, value, indent=0):
--- a/Code/autopkg
+++ b/Code/autopkg
@@ -35,6 +35,7 @@ except:
     print 'WARNING: importing plistlib as FoundationPlist'
     import plistlib as FoundationPlist
 
+import plistlib
 import pprint
 import re
 import shutil
@@ -49,6 +50,7 @@ from autopkglib import get_pref, get_all_prefs
 from autopkglib import set_pref, PreferenceError
 from autopkglib import AutoPackagerError, AutoPackager
 from autopkglib import get_processor, processor_names
+from autopkglib import extract_processor_name_with_recipe_identifier
 from autopkglib import get_autopkg_version, version_equal_or_greater
 from autopkglib import find_recipe_by_identifier, \
                        get_identifier
@@ -57,9 +59,9 @@ from autopkglib import find_recipe_by_identifier, \
 def log(msg, error=False):
     '''Message logger, prints to stdout/stderr.'''
     if error:
-        print >> sys.stderr, msg
+        print >> sys.stderr, msg.encode('UTF-8')
     else:
-        print >> sys.stdout, msg
+        print >> sys.stdout, msg.encode('UTF-8')
 
 
 def log_err(msg):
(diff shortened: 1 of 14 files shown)


gregneagle:
OK, that's a good start. Have you checked to see if there are any other situations that result in a exit result == 1?

timsutton:
I've found `1` happen in other scenarios as well, that's why I recommend (had just posted a [comment](https://github.com/autopkg/autopkg/issues/296#issuecomment-232736730) as Nick opened this PR) we could pick a different number for this specific case.

nmcspadden:
What would your preferred exit code be for 'some recipe failed'?

gregneagle:
None of these:
http://www.tldp.org/LDP/abs/html/exitcodes.html

nmcspadden:
So, 3 then?

timsutton:
At one point MagerValp pointed me to a good set of them, but I just found the Python `os` module reference to have a decent list of typical user-defined ones: https://docs.python.org/2/library/os.html#os.EX_OK

My understanding is that there's a precedent of starting at 64, but presumably you'd also use some of those things for other issues like wrong command arguments, file access, etc.

gregneagle:
Propose it. See if people object.

nmcspadden:
Based on @timsutton's suggestion, I think https://docs.python.org/2/library/os.html#os.EX_SOFTWARE is an appropriate choice to use, as there was an internal failure by a recipe failing to run properly.

gregneagle:
Or os.EX_CANTCREAT because the recipe failed to create the desired output file..

timsutton:
I'm not really arguing it should be one of those specific ones in that list, either - we've got plenty of "room" for more codes above the ones conveniently defined in `os`. If we see it being useful to be able to specifically distinguish this scenario of having a successful run but with failures, we can also pick a code that we won't ever use anywhere else.

natewalck:
I like the idea of knowing how many recipes failed just from the exit code. Maybe 201 = 1 recipe failed, 202 = 2 failed, etc?

Trying to get it out of range of standard exit codes, but that could get near 255+ then.

I like -1 for autopkg died and 0 for run was successfully.

nmcspadden:
To be honest, I really don't particularly care either way.  I don't really care what exit code it gets. I just want to distinguish "autopkg crashed" vs. literally any other outcome. If we want to distinguish different kinds of crashes, that's fine too, but that's not what I want to fix right now - that's something we can work on in a separate issue. 

The immediate issue is that there's no exit code distinction between "I crashed" and "I failed exactly 1 recipe", and that's what needs to be solved because otherwise wrapping `autopkg run` becomes a real challenge.

As per https://github.com/autopkg/autopkg/issues/296#issuecomment-232748783 as well, I also don't mind the return code being 'number of recipes that failed' so long as every occurrence of autopkg crashing is caught and returns something other than 1.  That seems like it's a bit harder and longer to solve, and in the meantime, the original problem still exists.  Can we guarantee that we can catch autopkg faulting and return -1 in each case?

timsutton:
I think if you want more details on how many recipes failed then you go get a list of failures from the report plist.

I also don't really care what exit code it gets, but I think if we're going to talk about modifying how AutoPkg reports exit status we should make a change we can stick with (since this is something consumed and relied upon in wrapper scripts, etc.).

Right now you'll get a `1` on pretty much any crash as far as I know, which is the catch-all error. In the future we could use some of the other more specific user errors defined in `os` if we want to be good UNIX citizens. Those ones go up to around 78.

What if we just round up and say 80? We can reserve this to be specific for a run completion but with one or more recipe failures. I think we should put this in as a commented constant somewhere rather than burying it in the function, though.

nmcspadden:
Would you prefer that constant be defined in autopkglib, or is it safe enough in autopkg itself?

timsutton:
Eventually maybe more of `autopkg` can be refactored elsewhere to make it more manageable, but for now all the logic of the tool itself is pretty self-contained, so might as well put error code(s) in `autopkg` too for the time being.

timsutton:
Working here. I noticed that if a recipe can't be found, and AutoPkg falls back to an interactive search prompt (and I say no), that must also be treated like a failure and results in the same exit code.

It's simplest to just assume that's also a recipe failure (you can argue it is, since the search operation for this recipe doesn't start until we actually "arrive" at this recipe in the run).

PR #303, opened 2016-08-03
Allowing for symlink permissions to be modified

n8felton:
This small change allows for pkgs created with the PkgCreator processor to contain symlinks, otherwise autopkgserver would fail claiming that the symlink'd file "could not be found"

### DIFF
--- a/Code/autopkgserver/packager.py
+++ b/Code/autopkgserver/packager.py
@@ -372,7 +372,7 @@ class Packager(object):
                     self.log.info("Setting mode of %s to %s" % (
                         entry.path,
                         str(entry.mode)))
-                    os.chmod(chownpath, int(entry.mode, 8))
+                    os.lchmod(chownpath, int(entry.mode, 8))
             else:
                 for (dirpath,
                      dirnames,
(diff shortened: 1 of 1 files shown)


gregneagle:
I wonder if line 375 should have the same change.

gregneagle:
I wonder why this hasn't been an issue for other recipes that use the Symlinker processor and create a package.

https://github.com/autopkg/foigus-recipes/tree/master/ZevrixSolutions has a few;
https://github.com/autopkg/jleggat-recipes/blob/master/MacVim/MacVim.pkg.recipe is another, and there's this one:
https://github.com/autopkg/recipes/blob/master/AutoPkg/AutoPkgGitMaster.pkg.recipe

n8felton:
The particular issue I'm seeing is that I am trying to symlink to a symlink, and because the code doesn't currently use `lchmod`, the PkgCreator is failing with `No such file or directory`.

From what I can tell with the other recipes with the Symlinker processor, they are all symlinking to hard links, not other symlinks.

n8felton:
I'd like to look at this PR again. It appears to still be able to merge in it's current form, and would allow me to not have to keep patching my local install.

nmcspadden:
I agree that line 375 should also be changed. The behavior should be consistent for each touching of the symlink.

n8felton:
Well, that'd didn't go as planned. I can't resolve the conflict or see the change request now. ¯\_(ツ)_/¯

nmcspadden:
Please rebase this and re-test if you want this to be evaluated.

PR #307, opened 2016-09-28
enable glob'd source path for payload to unpack

arubdesu:
Adds checking to bail if more than one result, otherwise relies on ditto to fail if path not valid/found. Updated variable description to reflect recommended hygiene if leveraging glob in recipe


arubdesu:
(Feel free to adjust the description I added, it's probably confusing)

arubdesu:
Er, feel free to let this die on the vine, I figured out how to do what I needed with FileFinder...

homebysix:
💀

Issue #313, opened 2016-10-26
Copier mounts dmg unnecessarily instead of copying


homebysix:
If I try to copy a dmg from one place to another using Copier, it mounts the dmg and fails due to permissions on an enclosed folder.

Here are the arguments that result in failure:

```
<dict>
    <key>Processor</key>
    <string>Copier</string>
    <key>Arguments</key>
    <dict>
        <key>source_path</key>
        <string>%RECIPE_CACHE_DIR%/downloads/ParallelsDesktop-%MAJOR_VERSION%.dmg</string>
        <key>destination_path</key>
        <string>%pkgroot%/private/tmp/ParallelsDesktopAutoDeploy.pkg/Parallels/ParallelsDesktop-%version%.dmg</string>
        <key>overwrite</key>
        <string>true</string>
    </dict>
</dict>
```

And here's the error from verbose output:

```
Copier
Copier: Mounted disk image ~/Library/AutoPkg/Cache/com.example.pkg.ParallelsDesktopAutoDeploy/downloads/ParallelsDesktop-12.dmg
Can't copy /private/tmp/dmg.RIWw4F/ to ~/Library/AutoPkg/Cache/com.example.pkg.ParallelsDesktopAutoDeploy/payload/private/tmp/ParallelsDesktopAutoDeploy.pkg/Parallels/ParallelsDesktop-12.0.2.dmg: [(u'/private/tmp/dmg.RIWw4F/.Trashes', u'~/Library/AutoPkg/Cache/com.example.pkg.ParallelsDesktopAutoDeploy/payload/private/tmp/ParallelsDesktopAutoDeploy.pkg/Parallels/ParallelsDesktop-12.0.2.dmg/.Trashes', "[Errno 13] Permission denied: '/private/tmp/dmg.RIWw4F/.Trashes'")]
Failed.
```

I found a workaround, which is to add a wildcard into the extension of the source_path. Perhaps this obscures from Copier the fact that it's dealing with a dmg, preventing the unnecessary mounting.

Here is the modified source_path:

```
        <key>source_path</key>
        <string>%RECIPE_CACHE_DIR%/downloads/ParallelsDesktop-%MAJOR_VERSION%.d?g</string>
```

In any case, if you're just copying a dmg from one place to another, mounting the dmg isn't desired or necessary.

gregneagle:
[...]

PR #314, opened 2016-10-26
CopierFix

nmcspadden:
https://github.com/autopkg/autopkg/issues/313

### DIFF
--- a/Code/autopkglib/Copier.py
+++ b/Code/autopkglib/Copier.py
@@ -85,7 +85,7 @@ class Copier(DmgMounter):
         # Check if we're trying to copy something inside a dmg.
         (dmg_path, dmg, dmg_source_path) = self.parsePathForDMG(source_path)
         try:
-            if dmg:
+            if dmg and dmg_source_path:
                 # Mount dmg and copy path inside.
                 mount_point = self.mount(dmg_path)
                 source_path = os.path.join(mount_point, dmg_source_path)
@@ -110,7 +110,7 @@ class Copier(DmgMounter):
             self.copy(matched_source_path, self.env['destination_path'],
                       overwrite=self.env.get("overwrite"))
         finally:
-            if dmg:
+            if dmg and dmg_source_path:
                 self.unmount(dmg_path)
 
 


gregneagle:
While that sounds like a good approach, didn't we find that code elsewhere called os.path.normpath()  which had the effect of stripping trailing slashes?

os.path.normpath('/path/to/some.dmg/') returns '/path/to/some.dmg'

gregneagle:
```python
#!/usr/bin/python

import os


class DmgMounter(object):
    """Base class for Processors that need to mount disk images."""

    DMG_EXTENSIONS = ['.dmg', '.iso', '.DMG', '.ISO']

    #pylint: disable=invalid-name
    def parsePathForDMG(self, pathname):
        """Helper method for working with paths that reference something
        inside a disk image"""
        dmg_path = os.path.normpath(pathname)
        dmg_source_path = ''
        while dmg_path != '/':
            for extension in self.DMG_EXTENSIONS:
                found_ext = os.path.splitext(dmg_path)[1]
                if found_ext == extension:
                    # we return the found extension with a trailing slash
                    # to match the previous implementation
                    return dmg_path, found_ext + '/', dmg_source_path
            dmg_path, tail = os.path.split(dmg_path)
            dmg_source_path = os.path.normpath(os.path.join(
                tail, dmg_source_path))
        # no disk image in path
        return pathname, '', ''
    #pylint: enable=invalid-name


dm = DmgMounter()

test_paths = (
    '/path/to/foo.dmg',
    '/path/to/foo.dmg/',
    '/path/to/foo.dmg/Applications/Foo.app/Contents/Info.plist',
    '/path/to/.dmg.tmp/foo.dmg',
    '/path/to/.dmg.tmp/foo.dmg/Applications/Foo.app/Contents/Info.plist',
    '/path/that/has/no/.dmg',
    '/path/that/has/no/.dmg/anywhere'
)

for pathname in test_paths:
    print "Input:   ", pathname
    print "Output:  ", dm.parsePathForDMG(pathname)
    print
```

Output of note:

```
Input:    /path/to/foo.dmg
Output:   ('/path/to/foo.dmg', '.dmg/', '')

Input:    /path/to/foo.dmg/
Output:   ('/path/to/foo.dmg', '.dmg/', '')
```

Is that a problem for the intended solution? Perhaps not, unless the recipe author intends to copy the entire contents of a diskimage somewhere (and not the disk image itself)

nmcspadden:
`parsePathForDMG` only populates the "dmg_source_path" variable if there's content afterwards, it seems; the return value in that tuple for `parsePathForDMG()` is empty, which is why checking for both `dmg` and `dmg_source_path` seem to do the right thing.

It does mean, as you pointed out, that it's now more difficult to copy the entire contents of a diskimage out.  I'm not sure how best to accommodate that, but a brief spot check in the repos doesn't show use cases for that.

gregneagle:
This is why I talked about enumerating and discussing the various possible use cases. We need to either handle them or document that they are not supported.

oucsaw:
(Commenting here, as this seems more appropriate than #313 or opening
another issue)

We've run into an issue with Copier.py mounting, instead of copying a .iso file.

After installing autopkg 1.0.1, one of our local recipes which uses:

```
    <dict>
        <key>Processor</key>
        <string>Copier</string>
        <key>Arguments</key>
        <dict>
            <key>source_path</key>
            <string>/path/to/file/Package.iso</string>
            <key>destination_path</key>
            <string>%RECIPE_CACHE_DIR%/%NAME%/private/tmp</string>
        </dict>
    </dict>
```

where what we are trying to achieve is broadly:

```
   cp /path/to/file/Package.iso %RECIPE_CACHE_DIR%/%NAME%/private/tmp
```

is now failing, complaining that:
```
Error in recpieid: Processor: Copier: Error: /path/to/file/Package.iso not mounted
```
(this is because we don't have sufficient permissions for hdiutil to
mount the .iso file, but this is an aside ...)

The change seems to be that `DmgMounter.parsePathForDMG` returns:
```
('/path/to/file/Package.iso', '.iso/', '')
```
whereas it used to return:
```
('/path/to/file/Package.iso', '', '')
```

Is this a correct usage of the Copier processor ? (i.e. should we be
using something else to copy .iso and .dmg files) 

Also, is there a document that let people know the expected/desired behavior
of parsePathForDMG ? (It is not clear at the moment, and thus is seems
wrong to try to patch this look until we know what it should be doing).

I've had a few thoughts on alternatices to fixing this - a RawCopier that
doesn't try mount things, split the input to an optional source_dmg and source_path
but these seem to dodge the issue above a bit.

Thanks for any help!

homebysix:
@oucsaw You might be able to work around the issue by using a wildcard character, like so:

```
    <dict>
        <key>Processor</key>
        <string>Copier</string>
        <key>Arguments</key>
        <dict>
            <key>source_path</key>
            <string>/path/to/file/Package.i?o</string>
            <key>destination_path</key>
            <string>%RECIPE_CACHE_DIR%/%NAME%/private/tmp</string>
        </dict>
    </dict>
```

I've used the same approach for dmgs and it worked well.

Regardless, I'm looking forward to #313 being solved.

oucsaw:
@homebysix: thanks for this! We've put this in place (hopefully temporarily) things are now  working

gregneagle:
Commit 5871fe4 was reverted today. DmgMounter returns to its previous behavior. Does that change/eliminate the need for this change?

nmcspadden:
Yes, this can be safely discarded.

PR #310, opened 2016-10-19
Removing temporary file when CHECK_FILESIZE_ONLY is true and download_changed is false

fuzzylogiq:
I think this explains why foigus was seeing this issue https://groups.google.com/forum/#!topic/autopkg-discuss/z7_jMuioOJU/discussion and hopefully fixes it.

PR #318, opened 2016-11-04
Remove temporary download on file size match.

jessepeterson:
When `CHECK_FILESIZE_ONLY` is in use the `URLDownloader` processor does not remove the temporary downloaded file.


jessepeterson:
Looks like this is a duplicate of #310.

PR #336, opened 2017-01-24
Fix regression in behavior of DmgMounter

jessepeterson:
DmgMounter would return a dmg path when no files inside a dmg were found.


jessepeterson:
@nmcspadden pointed out #313 in which the original problem reported by @homebysix is what this PR intends to fix (copying a DMG and not the contents thereof).

jessepeterson:
Note per #314 this does not fix the trailing-slash case (where you'd *want* to copy the contents of a dmg).

gregneagle:
I reverted 5871fe4  -- DmgMounter reverts to previous behavior.

PR #338, opened 2017-02-02
Added FTP response checking to URLDownloader processor

overhacked:
I found that for **ftp://** URLs (like those used by HP in their product driver downloads), autopkg always reports the file as newly downloaded. This patch allows `URLDownloader` to at least grab the file size from the FTP 213 response code that curl elicits by default via a SIZE command.

I also check for FTP success (125 or 150) and failure (55x) and map it to HTTP 200 and 404 in case any future changes check for those values in `header['http_result_code']`.

### DIFF
--- a/Code/autopkglib/URLDownloader.py
+++ b/Code/autopkglib/URLDownloader.py
@@ -218,6 +218,23 @@ class URLDownloader(Processor):
                         header[fieldname] = part[1]
                     except IndexError:
                         header[fieldname] = ''
+                elif self.env["url"].startswith('ftp://'):
+                    part = info.split(None, 1)
+                    responsecode = part[0]
+                    if responsecode == '213':
+                        # This is the reply to curl's SIZE command on the file
+                        # We can map it to the HTTP content-length header
+                        try:
+                            header['content-length'] = part[1]
+                        except IndexError:
+                            pass
+                    elif responsecode.startswith('55'):
+                        header['http_result_code'] = '404'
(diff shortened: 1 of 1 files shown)


gregneagle:
Looks reasonable to me; I've called for some additional eyes on it.

gregneagle:
Please see https://groups.google.com/d/msg/autopkg-discuss/kdrvz2UR2eo/-Y-fbxI1CAAJ

timsutton:
It looks good to me as well. I didn't write any of URLDownloader's header parsing code so I'm much less familiar with it, but is it ok that `donewithheaders = True` is set only for the `150` or `125` response codes?

overhacked:
I think setting `donewithheaders = True` only on success follows the pattern of the pre-existing HTTP response-handling code. Curl exits quickly on failure, and there is also a catch-all `else` that will set `donewithheaders = True` when the while-loop hits a blank line (or end of input). The only difference with FTP is that there won't be any blank lines unless cURL changes its output format. End of input comes after the file transfer is complete and cURL closes stdout, so staying in the while-loop until then is basically the same as the current behavior, synchronously waiting on the process to exit.

overhacked:
I saw no objections on the Google Groups review thread. Could you consider merging this for the next release?

Apologies if there is a PR policy in place that I'm unaware of.

timsutton:
I don't have any objections, and think it's worth merging. There's no formal PR policy, don't worry, you didn't miss anything.

PR #341, opened 2017-02-20
CodeSignatureVerifier improvements

hjuutilainen:
This PR adds a number of new things and modernizations to CodeSignatureVerifier.

- Added `--strict` option to codesign command if on 10.11 or later. This is going to break some recipes but verifying with it is recommended by Apple in the code signing guide. The next new thing is going to help with possible problems:
- Added support for specifying additional arguments for `codesign`. This should help with situations where we need relax the default verification a bit. For example, we could now do this and add verification to Chrome recipe:

```
<dict>
	<key>Processor</key>
	<string>CodeSignatureVerifier</string>
	<key>Arguments</key>
	<dict>
		<key>input_path</key>
		<string>%pathname%/Google Chrome.app</string>
		<key>requirement</key>
		<string>(identifier "com.google.Chrome" or identifier "com.google.Chrome.canary") and certificate leaf = H"85cee8254216185620ddc8851c7a9fc4dfe120ef"</string>
		<key>codesign_additional_arguments</key>
		<array>
			<string>--no-strict</string>
		</array>
	</dict>
</dict>
```

- Removed functionality to use `expected_authority_names` key with codesign. This was already printing a warning message but I think it's time for it to go away completely.
- Removed the .app file extension checking and no longer require the input to be a specific file type. Only check for .pkg, .mpkg or .xip extensions and pass those to pkgutil, everything else should go to codesign.

And don't you dare merge this yet since I have only been running this on 10.12 machines. I need to go through and test all this with earlier system versions. But let the discussion begin.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,24 @@
-### [1.0.2](https://github.com/autopkg/autopkg/compare/v1.0.1...HEAD) (Unreleased)
+### [1.0.3](https://github.com/autopkg/autopkg/compare/v1.0.2...HEAD) (Unreleased)
+
+### [1.0.2](https://github.com/autopkg/autopkg/compare/v1.0.1...v1.0.2) (April 07, 2017)
+
+FIXES:
+
+- Trust info is ignored if it is ever present in a recipe which is not an override,
+  and warns if any such trust info is found (GH-334)
+- Fix a regression in PlistReader and handling disk images
+- PkgCopier can now mount a diskimage in the source path even if its extension is not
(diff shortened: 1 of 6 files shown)


arubdesu:
Per my comments in Slack, do we want to do either a less-than-16k file size check (to catch html 404 redirect/other error messages) or an executable (Mach-O bundle) file check around https://github.com/hjuutilainen/autopkg/blob/3cc892cc0f140622ea3b43cfbbb6bc51aa059c95/Code/autopkglib/CodeSignatureVerifier.py#L275 ?

PR #343, opened 2017-02-23
Improved "Invalid version" error reporting

magnusviri:
Added code that prints out the text that is causing "Invalid version" errors.


gregneagle:
Seems fine to me...

PR #352, opened 2017-05-28
MunkiImporter: add bundle types when making catalog db

timsutton:
I found that I needed to add this for an item where an installs array of a single `bundle` type was merged into `pkginfo` using `MunkiPkginfoMerger` prior to calling `MunkiImporter`.

It's not quite obvious to me why this wasn't needed originally in #216, but it seems like without it, the item never gets added to the pkgdb. Current the applications "table" only contains keys which are also valid for `bundle` types, so this seems safe enough, but would appreciate a sanity check.


timsutton:
I'm not sure why I held off merging this, probably just intended to come back to it again and forgot.

PR #357, opened 2017-07-03
Define filesystem when creating disk images

hjuutilainen:
Added default filesystem (Journaled HFS+) and support for overriding it to DmgCreator.

It seems that `hdiutil create` on  High Sierra currently creates disk images with APFS filesystem by default if the `-fs` argument is not used. This is probably not what the default should be for AutoPkg.

### DIFF
--- a/Code/autopkglib/DmgCreator.py
+++ b/Code/autopkglib/DmgCreator.py
@@ -24,6 +24,7 @@ from autopkglib import Processor, ProcessorError
 __all__ = ["DmgCreator"]
 
 DEFAULT_DMG_FORMAT = "UDZO"
+DEFAULT_DMG_FILESYSTEM = "Journaled HFS+"
 DEFAULT_ZLIB_LEVEL = 5
 
 class DmgCreator(Processor):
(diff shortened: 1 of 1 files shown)


timsutton:
As soon as I saw https://twitter.com/hjuutilainen/status/881207245320720385 I had the same thought. Thanks!

gregneagle:
Is Journaled HFS+ really the default filesystem for `hdiutil create` In macOS 10.12 and below?

timsutton:
Now that you mention it - seems like it would be weird to create Journaled HFS+ for DMGs by default.

On my 10.12.5 system I did `hdiutil create -srcfolder /Applications/Firefox.app ff.dmg`, and opened it in Disk Utility, and indeed it's not journaled.

screenshot 2017-07-03 22 00 36

hjuutilainen:
I somehow got journaled HFS+ when testing what's the default when creating images with Disk Utility but plain HFS+ is obviously the correct one to use. Will create a new PR and also add `Case-insensitive APFS` to the valid filesystems list because 10.12.5 hdiutil lists that.

Issue #355, opened 2017-06-24
TypeError when attempting to run recipe with missing parent


homebysix:
I'm on the master branch (1f0df1a), version 1.0.3.

When I attempt to run a recipe with a missing parent, I get the typical "Search GitHub" prompt, but no matter whether I enter `y` or `n`, a TypeError occurs.

```
$ autopkg version
1.0.3
```

```
$ autopkg run -v Postbox.pkg
Didn't find a recipe for com.github.gerardkok.download.Postbox.
Search GitHub AutoPkg repos for a com.github.gerardkok.download.Postbox recipe? [y/n]: n
Could not find parent recipe for Postbox.pkg
Traceback (most recent call last):
  File "/usr/local/bin/autopkg", line 2435, in <module>
    sys.exit(main(sys.argv))
  File "/usr/local/bin/autopkg", line 2429, in main
    exit(subcommands[verb]['function'](argv))
  File "/usr/local/bin/autopkg", line 1963, in run_recipes
    search_github=make_suggestions)
  File "/usr/local/bin/autopkg", line 405, in load_recipe
    elif 'ParentRecipeTrustInfo' in recipe:
TypeError: argument of type 'NoneType' is not iterable
```

```
$ autopkg run -v Postbox.pkg
Didn't find a recipe for com.github.gerardkok.download.Postbox.
Search GitHub AutoPkg repos for a com.github.gerardkok.download.Postbox recipe? [y/n]: y

Name                        Repo                  Path
----                        ----                  ----
Postbox.munki.recipe        gerardkok-recipes     Postbox/Postbox.munki.recipe
Postbox.download.recipe     gerardkok-recipes     Postbox/Postbox.download.recipe
Postbox.install.recipe      homebysix-recipes     Postbox/Postbox.install.recipe
Postbox.pkg.recipe          homebysix-recipes     Postbox/Postbox.pkg.recipe

To add a new recipe repo, use 'autopkg repo-add <repo name>'
Could not find parent recipe for Postbox.pkg
Traceback (most recent call last):
  File "/usr/local/bin/autopkg", line 2435, in <module>
    sys.exit(main(sys.argv))
  File "/usr/local/bin/autopkg", line 2429, in main
    exit(subcommands[verb]['function'](argv))
  File "/usr/local/bin/autopkg", line 1963, in run_recipes
    search_github=make_suggestions)
  File "/usr/local/bin/autopkg", line 405, in load_recipe
[...]

PR #363, opened 2017-07-12
Update autopkg

bochoven:
Add If statement to prevent stack trace when parent recipe does not exist (anymore)


bochoven:
The block stays well within 80 chars.
The if statement also has to cover

elif 'ParentRecipeTrustInfo' in recipe

timsutton:
Thanks @bochoven! I assume this resolves #355?

PR #368, opened 2017-08-22
Add repo as submodule when autopkg settings is versioned

ygini:
This change allow the autopkg working folder to be versioned by git.

The goal is to allow IT to manage the whole autopkg setup under git and so, repo should be added as submodule instead of simple clone.

Managing repo as submodule with a versioned autopkg config will allow a team to execute autopkg on multiple workstation and still relying on the same list of repo with the same validated version for each.

For example, if the IT team do its security job and check each repo update before accepting it, submodule can be used to track this validation.

### DIFF
--- a/Code/autopkg
+++ b/Code/autopkg
@@ -529,11 +529,25 @@ def get_recipe_repo(git_path):
     if not gitcmd:
         log_err("No git binary could be found!")
         return None
+    
+    # check if recipe_repo_dir is versionned with git or not
+    # if so, we should use submodule instead of clone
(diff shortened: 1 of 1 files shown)


ygini:
For information, I use this change on my own setup for almost a month without any issue.

gregneagle:
Does this change affect repo-update or repo-delete ?

ygini:
Hi

The update process is handled. Not the delete one, I've missed it. 

Don't merge the PR yet, I will submit ASAP the missing part.

ygini:
Support for delete added, and an issue found and fixed in the same time.

PR #373, opened 2017-09-06
New options for CodeSignatureVerifier

hjuutilainen:
This PR adds a number of new things and modernizations to CodeSignatureVerifier without changing current behavior too much.

- Added `strict_verification` variable to control whether to pass `--strict`, `--no-strict` or nothing at all to codesign. The default is to not pass anything unless specifically defined. This will make the strict verification an opt-in for recipe writers but I think it's better than breaking things. Using this in a recipe might look like this:

```
<dict>
	<key>Processor</key>
	<string>CodeSignatureVerifier</string>
	<key>Arguments</key>
	<dict>
		<key>input_path</key>
		<string>%pathname%/Google Chrome.app</string>
		<key>requirement</key>
		<string>(identifier "com.google.Chrome" or identifier "com.google.Chrome.beta" or identifier "com.google.Chrome.dev" or identifier "com.google.Chrome.canary") and (certificate leaf = H"85cee8254216185620ddc8851c7a9fc4dfe120ef" or certificate leaf = H"c9a99324ca3fcb23dbcc36bd5fd4f9753305130a")</string>
		<key>strict_verification</key>
		<false />
	</dict>
</dict>
```

Or one could replicate all the different verification steps the Chromium project is using: https://chromium.googlesource.com/chromium/src/+/lkgr/chrome/installer/mac/sign_app.sh.in

- Added `deep_verification` variable to control whether to pass `--deep` to codesign or not. Deep verification was previously on by default (and still is) but it can now be explicitly disabled.

- Added `codesign_additional_arguments` variable for specifying additional arguments for `codesign` tool.

- Removed the .app file extension checking and no longer require the input to be a specific file type. Only check for .pkg, .mpkg or .xip extensions and pass those to pkgutil, everything else should go to codesign.

The only breaking change in this PR should be this: Removed functionality to use `expected_authority_names` key with codesign. This was already printing a warning message but I think it's time for it to go away completely. The last time I checked, there was 5 recipes using this combo and they will break. I will personally find and fix these through pull requests.

### DIFF
--- a/Code/autopkglib/CodeSignatureVerifier.py
+++ b/Code/autopkglib/CodeSignatureVerifier.py
@@ -106,7 +106,7 @@ class CodeSignatureVerifier(DmgMounter):
         # Use --deep option in OS X 10.9.5 or later
         darwin_version = os.uname()[2]
         if StrictVersion(darwin_version) >= StrictVersion('13.4.0'):
-            if deep_verification is True:
+            if deep_verification:
                 self.output("Deep verification enabled...")
                 process.append("--deep")
             else:
@@ -116,12 +116,14 @@ class CodeSignatureVerifier(DmgMounter):
         if StrictVersion(darwin_version) >= StrictVersion('15.0'):
             if strict_verification is None:
                 self.output("Strict verification not defined. Using codesign defaults...")
-            elif strict_verification is True:
+            elif strict_verification:
                 self.output("Strict verification enabled...")
                 process.append("--strict")
-            elif strict_verification is False:
+            elif not strict_verification:
                 self.output("Strict verification disabled...")
                 process.append("--no-strict")
+            else:
+                self.output("Strict verification value type unknown. Using codesign defaults...")
 
         # Add additional arguments (if any).
         for argument in codesign_additional_arguments:
--- a/Code/autopkglib/CodeSignatureVerifier.py
+++ b/Code/autopkglib/CodeSignatureVerifier.py
@@ -26,7 +26,6 @@ from autopkglib.DmgMounter import DmgMounter
 
 __all__ = ["CodeSignatureVerifier"]
 
-RE_AUTHORITY_CODESIGN = re.compile(r'Authority=(?P<authority>.*)\n')
 RE_AUTHORITY_PKGUTIL = re.compile(r'\s+[1-9]+\. (?P<authority>.*)\n')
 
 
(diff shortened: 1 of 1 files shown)


gregneagle:
Your last pass at this sort of thing you warned _not_ to merge this because you hadn't tested it broadly enough. How about this one?

hjuutilainen:
I’ve been running these in production for a while now but only with 10.12 and 10.13. Gimme a day or two to run with older systems just to make sure. 

But the changes are now something that might be easier to merge and accept? I’m pretty happy with the new deep and strict implementation with boolean flags instead of blindly passing args to codesign.

Code/autopkglib/CodeSignatureVerifier.py
```diff
@@ -174,25 +209,16 @@ def process_app_bundle(self, path):
                 "DISABLE_CODE_SIGNATURE_VERIFICATION to a non-empty value.")
 
         if self.env.get('expected_authority_names', None):
-            self.output("WARNING: Using 'expected_authority_names' to verify .app "
-                        "bundles is deprecated and may be removed in a future "
-                        "AutoPkg release. Verifying .app bundles should use the "
+            self.output("ERROR: Using 'expected_authority_names' to verify code "
+                        "signature is deprecated. Recipes should use the "
```
gregneagle: 'deprecated' implies 'still works but you should stop using this'. Shouldn't this be "ERROR: Using 'expected_authority_names' to verify code signature is no longer supported..."

hjuutilainen: Yep, updated the wording.

hjuutilainen:
Tested these changes today with 10.7 and 10.11 using all of my .download recipes and specifically enabled strict verification for all of them. No issues related to these changes.

Code/autopkglib/CodeSignatureVerifier.py
gregneagle: So it still works, then? I thought you had removed support. If so, 'deprecated' is not the right word here.

hjuutilainen: I changed "deprecated" to "no longer supported" in cf9367c1e21736d6a5eb60eba73f8f9129471b36. The error is only raised when verifying code signatures (with `codesign`), not installer package signatures (`pkgutil`). I need to write a better description for `expected_authority_names` in input variables section.

gregneagle:
Are _you_ happy with the code as it stands? If so, I'll merge.

hjuutilainen:
Yes, I’m happy with it.

homebysix:
> The only breaking change in this PR should be this: Removed functionality to use `expected_authority_names` key with codesign.

Is it safe to remove mention of this key from the wiki, then? I'd be happy to do the edits, if so. This change slipped by me.

macmule:
We have 10 recipes within our repo using "expected_authority_names" alone (  https://github.com/autopkg/dataJAR-recipes/search?utf8=✓&q=expected_authority_names&type=).

We'll update our recipes

hjuutilainen:
Using expected_authority_names was removed for everything using `codesign` in the background. Verifying installer package (.pkg) signatures uses `pkgutil` tool and using expected_authority_names in those recipes is still valid and supported.

macmule:
@hjuutilainen tah!

Issue #376, opened 2017-10-04
Please fork my repo


aschwanb:
https://github.com/ITS-Unibas/autopkg-recipes

A repo with recipes for mainly edu focused software maintained by the IT services at the university of Basel.

If possible, we would like to keep using this repo as a team and update/add recipes we need. Could you name the repo `its-unibas-recipes` to reflect this.

Thank you very much.

hjuutilainen:
Thank you for these. I'll fork this soon but I have a couple of notes and minor nitpicks for the recipes:

- Cleanup whitespace to be consistent (use either tabs or spaces). This will make the recipes easier to read when viewing in browser.
- In download recipes, put EndOfCheckPhase processor right after the URLDownloader etc. and before any extraction or signature verification. This will speed up running autopkg in check mode since it doesn't need to extract and verify if nothing has changed.
- In Karabiner.download, I would personally move the Versioner and MunkiInstallsItemsCreator to the Munki recipe. I would also check the signature of the installer package only and not extract the package at all.
- Does F5Transkript.download really need a custom processor to get the download URL? You can of course use a custom processor but in this case I think you could use URLTextSearcher to extract the needed bits from the download page.

aschwanb:
Hey @hjuutilainen 

I updated the recipes according to your suggestions but I think we need to keep the custom processor for F5Transkript.download. The reason for it is that we get an escaped URL from the download page.

We basically need is a way to string replace from `&amp;` to `&`. But I don't know if this could be done with URLDownloader or URLTextSearcher. Do you know if this could be done with the current processors?

aschwanb:
Update: I opened a pull request with the proposed change. See #380

hjuutilainen:
I think we've had a few discussions about unescaping URLs before but if I recall it correctly there were always other issues caused by this. Let's discuss this more in #380.

For the F5Transkript.download recipe, you could use a single URLTextSearcher to capture multiple strings (with Python RE named groups) and then compose the download URL from the pieces but it's not pretty:

```
<dict>
	<key>Processor</key>
	<string>URLTextSearcher</string>
	<key>Arguments</key>
	<dict>
		<key>url</key>
		<string>https://www.audiotranskription.de/downloads.html</string>
		<key>re_pattern</key>
[...]

PR #380, opened 2017-10-10
Unescape URLs found with URLTextSearcher

aschwanb:
### Proposed Change
This pull request adds a method to handle developers that only provide escaped urls on their download page. As an example see [this download recipe.](https://github.com/ITS-Unibas/autopkg-recipes/blob/F5/F5transkript/F5transkript.download.recipe)

Output now: `{'Output': {'match': '/audot/downloadfile.php?k=1&amp;d=13&amp;l=de&amp;c=2d9cf5ca79'}}`

Output after the change to `URLTextSearcher`: `{'Output': {'match': u'/audot/downloadfile.php?k=1&d=13&l=de&c=2d9cf5ca79'}}`

### Reasons for change:
We use a [a custom URLProvider](https://github.com/ITS-Unibas/autopkg-recipes/blob/master/F5transkript/F5transkriptURLProvider.py) in our autopkg recipe for [F5transkript](https://github.com/ITS-Unibas/autopkg-recipes/blob/master/F5transkript/F5transkriptURLProvider.py). The reason for this being that the developer only publishes an unescaped URL on the download page and that there is not currently a way to unescape urls with curl - as far as I know.

I would like to add our repo to the autopkg organisation (See issue https://github.com/autopkg/autopkg/issues/376) and this pull request is a possible way to get rid of our custom URLProvider.

I know that such a change because of one piece of software is probably undesired - also, there might be a simpler way to get the correct url. But maybe there are others out there who face the same problem and would benefit from such a change as well.

I'm putting this up for discussion.

### DIFF
--- a/Code/autopkglib/URLTextSearcher.py
+++ b/Code/autopkglib/URLTextSearcher.py
@@ -18,6 +18,7 @@
 
 import re
 import subprocess
+import HTMLParser
 
 from autopkglib import Processor, ProcessorError
 
@@ -104,8 +105,11 @@ class URLTextSearcher(Processor):
         if not match:
             raise ProcessorError('No match found on URL: %s' % url)
 
-        # return the last matched group with the dict of named groups
-        return (match.group(match.lastindex or 0), match.groupdict(), )
+        # Unescape last matched group to get rid of html escape characters
+        unescaped_url = HTMLParser.HTMLParser().unescape(match.group(match.lastindex or 0))
+
+        # return the last matched group as unescaped url with the dict of named groups
+        return (unescaped_url, match.groupdict(), )
 
     def main(self):
         output_var_name = self.env['result_output_var_name']


aschwanb:
I've been thinking about this issue a little more. Changing the default behaviour of this processor is probably not the best solution as we don't know what recipes will break.

There is however the possibility of creating a new input variable. We could call it `unescape_url`. It would not be required, be set to `False` by default and if set to `True` would run 
```HTMLParser.HTMLParser().unescape(match.group(match.lastindex or 0))```
bevor returning the value.

Would that be an acceptable solution?

gregneagle:
Just want to point out that even if this change were accepted, other people could not use any recipes that depended on this change until there was a release incorporating it, or were confident enough to use development code. That might lead to more support issues.

aschwanb:
@gregneagle Very true.

I proposed this change because I thought that other people might have similar issues they were trying to solve. As it stands it's probably not worth a change in one of autopkgs processors.

I might get back to this sometime, but for now I'll close this pull request.

PR #385, opened 2017-11-21
Discard temp file after matching file sizes

orbsmiv:
Ensure that `temporary_file` is removed if URLdownloader determines that the file hasn't changed by matching the new and existing file sizes.


orbsmiv:
Sorry – duplicate of #310.

PR #386, opened 2017-11-21
Add optional curl options to URLTextSearcher

orbsmiv:
Further to #384, also add support for additional curl options to URLTextSearcher.


gregneagle:
Looks OK to me.

PR #401, opened 2018-01-24
Use `os.makedirs` instead of `os.mkdir` for all core processors.

sheagcraig:
This PR is to replace the use of `os.mkdir` with `os.makedirs` for three of the core processors:
- FlatPkgUnpacker
- PkgPayloadUnpacker
- PkgRootCreator

Existing processors that already use `os.makedirs` are:
- MunkiImporter
- AppPkgCreator
- PkgExtractor
- URLDownloader
- Unarchiver

The thinking being that some of these processors will create intermediary directories if so instructed, while others won't. For the sake of consistency, it makes sense to use `os.makedirs` for all of the above.

`os.makedirs` has the same signature as `os.mkdir`, so there shouldn't be any issues.

### DIFF
--- a/Code/autopkglib/FlatPkgUnpacker.py
+++ b/Code/autopkglib/FlatPkgUnpacker.py
@@ -74,7 +74,7 @@ class FlatPkgUnpacker(DmgMounter):
         # Create the directory if needed.
         if not os.path.exists(self.env['destination_path']):
             try:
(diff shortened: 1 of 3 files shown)


hjuutilainen:
Thank you for the PR. Merging right away without further discussion since the chances are small and definitely needed.

Issue #404, opened 2018-02-02
GitHubReleaseInfoProvider unauthorized request HTTP 403


tcinbis:
Hi everyone!
We are trying to use Travis as CI for our recipes and get notifications once a recipe breaks.
With the `.travis.yml` file almost everything works, except recipes which are using the `GitHubReleasesInfoProvider`. 

The following error occures:
```
GitHubReleasesInfoProvider
{'Input': {'github_repo': u'ANTS-Framework/ants'}}
API error: HTTP Error 403: Forbidden
Unexpected GitHub API status code 403.
Failed.
Receipt written to /Users/travis/Library/AutoPkg/Cache/com.github.its-unibas.munki.ants_client/receipts/com.github.its-unibas.munki-receipt-20180126-134713.plist
The following recipes failed:
    com.github.its-unibas.munki.Anaconda
        Error in com.github.its-unibas.munki.Anaconda: Processor: MunkiImporter: Error: creating pkginfo for /Users/travis/Library/AutoPkg/Cache/com.github.its-unibas.munki.Anaconda/downloads/Anaconda2-5.0.1.sh failed: /Users/travis/Library/AutoPkg/Cache/com.github.its-unibas.munki.Anaconda/downloads/Anaconda2-5.0.1.sh is not a valid installer item!
    com.github.its-unibas.munki.ants_client
        Error in com.github.its-unibas.munki.ants_client: Processor: GitHubReleasesInfoProvider: Error: Unexpected GitHub API status code 403.
```
I checked the source code and saw that by default no token is used to authenticate. Maybe it helps to check for a token file by default?
I could create such a pull request in case this could fix the problem. Or maybe it is something travis related?

Travis config file:
```
os: osx
install:
- curl -LJO https://github.com/autopkg/autopkg/releases/download/v1.0.3/autopkg-1.0.3.pkg
- curl -LJO https://github.com/munki/munki/releases/download/v3.1.1/munkitools-3.1.1.3447.pkg
- sudo installer -verbose -pkg "autopkg-1.0.3.pkg" -target /
- sudo installer -verbose -pkg "munkitools-3.1.1.3447.pkg" -target /
- autopkg repo-add https://github.com/autopkg/its-unibas-recipes
- autopkg repo-add autopkg/recipes
- autopkg repo-add autopkg/hjuutilainen-recipes
- autopkg repo-add autopkg/homebysix-recipes
- autopkg repo-add autopkg/rtrouton-recipes
[...]

PR #407, opened 2018-02-05
Implemented new preference to support github tokens

tcinbis:
You can now add the key "GITHUB_TOKEN" to the autopkg default
preferences to make authenticated requests for all github related api
calls. 

In case there is no token specified, it will fallback to
unauthenticated request as before. The token should only have public
scope, because it is generally accessible on your computer!

This closes issue #404

### DIFF
--- a/Code/autopkglib/github/__init__.py
+++ b/Code/autopkglib/github/__init__.py
@@ -22,7 +22,7 @@ import urllib2
 
 from base64 import b64encode
 from getpass import getpass
-from pprint import pprint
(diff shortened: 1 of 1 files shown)


gregneagle:
Would definitely like @timsutton's comments. This change would also need some updated documentation .

tcinbis:
In case the PR will be approved I can also update the corresponding documentation.

tcinbis:
Because my changes got already merged with #409 we can close this PR.

PR #409, opened 2018-02-27
Use curl for GitHub API calls instead of urllib2

hjuutilainen:
These changes swap out urllib2 for curl when making GitHub API calls. I tried to keep changes to existing behaviour at minimum but some things are different:

- When using curl, we can't easily get the HTTP status code. I therefore removed the status code from the `call_api()` return value and just return the json response.
- When doing `autopkg search -t`, we're not trying to create a new token over the GH API. Instead just display a link where a token can be manually created and wait for user to copy/paste it back in.

This PR doesn't include any changes to the token behaviour, that will be a separate PR.

### DIFF
--- a/Code/autopkg
+++ b/Code/autopkg
@@ -820,18 +820,22 @@ def do_gh_code_search(query, use_token=False):
     if use_token:
         gh_session.setup_token()
     # Do the search, including text match metadata
-    results = gh_session.call_api(
+    (results, code) = gh_session.call_api(
         "/search/code",
         query=query,
         accept="application/vnd.github.v3.text-match+json")
     
-    gh_message = results.get("message", None)
-    gh_documentation_url = results.get("documentation_url", None)
-    if gh_message:
-        log_err(gh_message)
-        log_err(gh_documentation_url)
-    
-    if results is None:
+    if code == 403:
+        log_err(
+                "You've probably hit the GitHub's search rate limit, officially 5 "
+                "requests per minute.\n")
+        if results:
+            log_err("Server response follows:\n")
+            log_err(results.get("message", None))
+            log_err(results.get("documentation_url", None))
+            
+        return None
+    if results is None or code is None:
         log_err("A GitHub API error occurred!")
         return None
     return results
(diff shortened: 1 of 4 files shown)


gregneagle:
I'm curious why you say you can't easily get the http status code. Back when Munki was using curl, this code did get http status. Maybe it can help?
https://github.com/munki/munki/blob/Munki1/code/client/munkilib/fetch.py#L91-L427

gregneagle:
This code is similar: https://github.com/wdas/reposado/blob/master/code/repo_sync#L265-L453

hjuutilainen:
I’ll take a look at those if there’s something we can reuse. What I meant with easily was not having to write a few hundred lines of new code for this issue.

gregneagle:
...which is why I pointed you at some "tested"/not-new code... :-)

gregneagle:
I'll leave it to you decide if the need for/lack of the http status code is important enough to deal with at this time, and trust your decision in this. Feel free to merge the code when you are happy/confident.

macmule:
Not sure if this is of any help, it's a snippet from one of our things & will get the HTTP code

```
    curl_args = ['/usr/bin/curl', '-s', '-w', ' %{http_code} %{url_effective}', \
                                             '-d', jwt_complete, ENVQUERY_URL,]
    try:
        curl_output = subprocess.check_output(curl_args)
    except (subprocess.CalledProcessError, NameError) as error_msg:
        print 'ERROR: lookup with: %s' % error_msg
        sys.exit(1)

    try:
        return_code = curl_output.split()[0]
        end_url = curl_output.split()[1]
    except (ValueError, IndexError):
        print 'ERROR: invalid response: %s' % curl_output
        sys.exit(1)

    if end_url != SOME_URL:
        print 'ERROR: URL redirected: %s' % curl_output
        sys.exit(1)
    else:
        pass

    if return_code != '204':
        print 'ERROR: error response code: %s' % return_code
        sys.exit(1)
    else:
        pass
```

gregneagle:
So this includes https://github.com/autopkg/autopkg/pull/407 ?

We should get a 1.0.4 release candidate out ASAP since this affects a great number of autopkg users.

hjuutilainen:
Yes, #407 is included.

hjuutilainen:
I’ve run all of my own AutoPkg workflows with a couple of hundred recipes using this code and had no problems. I think this is good to go.

gregneagle:
Merge it!

hjuutilainen:
Merged now.

PR #420, opened 2018-06-02
Introduce the preprocessor MunkiNormalizePath

ygini:
MunkiNormalizePath is aim to be used with --pre to standardize imported item name and subpath
It will update NAME and MUNKI_REPO_SUBDIR variable to replace all non alphanumeric char with an underscore and will also convert everything to lowercase

Using it shouldn't break correclty written recipes. However, some has been written to use NAME everywhere, even on non-varaible things.

When deploying this preprocessor, be sure to check that all your recipes are still working. If there is any issue, send a PR to the recipe maintainer to avoid using NAME for non variable things.

### DIFF
--- a/Code/autopkglib/MunkiNormalizePath.py
+++ b/Code/autopkglib/MunkiNormalizePath.py
@@ -1,0 +1,64 @@
+#!/usr/bin/env python
+#
+# Copyright 2018 Yoann Gini
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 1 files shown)


gregneagle:
Since you can use processors that aren't part of AutoPkg core, I'd think it would make more sense for you to have this in a recipe repo _you_ are the owner of so you can maintain it as you'd like.

I'm not sure I really understand the need for this processor -- NAME and MUNKI_REPO_SUBDIR are under the control of the recipe runner -- the person can set these to whatever they like -- if they want all lowercase and no non-alphanumerics they can just do that.

ygini:
Goal is simply to avoid manual edition of every single recipe you trust, it's just a quick win option.

Regarding main repo or my repo, I've no opinions on this.

PR #424, opened 2018-07-03
Removed bundle as in option the description of pkgtype since it is no…

grahampugh:
… longer valid

### DIFF
--- a/Code/autopkgserver/autopkgserver
+++ b/Code/autopkgserver/autopkgserver
@@ -44,7 +44,7 @@ Request format:
     pkgdir          The directory where the pkg will be created, must be a
                     directory owned by the calling user.
 
-    pkgname         The file or bundle name of the pkg, without pkg extension.
+    pkgname         The file name of the pkg, without pkg extension.
 
     pkgtype         The package type. "flat" is the only valid value.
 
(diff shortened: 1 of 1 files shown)


Code/autopkgserver/autopkgserver
```diff
@@ -46,7 +46,7 @@ Request format:
 
     pkgname         The file or bundle name of the pkg, without pkg extension.
```
gregneagle: Might want to remove other occurrences of "bundle", like here

Code/autopkgserver/autopkgserver
```diff
@@ -46,7 +46,7 @@ Request format:
...
 
-    pkgtype         The package type, "flat" or "bundle".
+    pkgtype         The package type. "flat" is the only valid value.
```
gregneagle: Maybe `"flat" is currently the only valid value.` to deflect questions about why pkgtype _exists_ if it has only one valid value...

grahampugh:
I've removed the other references in `autopkgserver` and updated the commit. 

There are other references that may be relevant in `packager.py`, `PkgPayloadUnpacker.py`, `PkgInfoCreator.py` and `PkgExtractor.py`, but I'm not confident enough to know whether these are still required for legacy reasons.

gregneagle:
"There are other references that may be relevant in packager.py, PkgPayloadUnpacker.py, PkgInfoCreator.py and PkgExtractor.py, but I'm not confident enough to know whether these are still required for legacy reasons."

Certainly there may still be bundle-style packages that AutoPkg may have to work with, even if it doesn't _create_ them. Adobe CCP output is bundle-style packages, for example.

PR #442, opened 2019-01-09
Added RegexReplace processor

tcinbis:
This processor takes a string and a pattern and will replace all occurrences
of the pattern in the given string.

The need for this kind of processor aroused while updating the [Geneious.download.recipe](https://github.com/autopkg/n8felton-recipes/blob/master/Geneious/Geneious.download.recipe) to version 2019. Geneious stores their version information in a dedicated text file, but to download the dmg file a `-` needs to replaced with a `_`.
`Geneious_Prime-mac64_2019_0_4_with_jre.dmg`-> `GGeneious_Prime_mac64_2019_0_4_with_jre.dmg`

### DIFF
--- a/Code/autopkglib/RegexReplace.py
+++ b/Code/autopkglib/RegexReplace.py
@@ -1,0 +1,60 @@
+#!/usr/bin/env python
+
+from autopkglib import Processor, ProcessorError
+import re
+
+__all__ = ["RegexReplace"]
+
+
+class RegexReplace(Processor):
+    '''Replaces all occurrences of a given pattern in the passed string.'''
+
+    input_variables = {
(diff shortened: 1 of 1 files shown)


tcinbis:
Any suggestions or further improvements required?
@hjuutilainen @gregneagle

gregneagle:
So far exactly one recipe has needed this -- that recipe can include it with itself.
I'm not _opposed_ to this, but we're not doing a great job of keeping up with the support and maintenance of the existing tools -- adding more processors for the core maintainers to support/maintain won't help in that regard. Again: this is not a _no_. Perhaps the other maintainers feel differently...

PR #452, opened 2019-03-06
Added option to Stop Processing if the URLDownloader reports the download did not change

grahampugh:
This PR is to give users the option to override recipes to stop processing after `URLDownloader` in the case that the download has not changed. 

There are two parts to this change:

1. This change adds a default of `FALSEPREDICATE` to the `StopProcessingIf` processor. This means that the processor can be added to recipes with no arguments in a recipe, so that users could override the `predicate` value. The default value of `FALSEPREDICATE` always evaluates to `False`, so the processor will do nothing without supplying an argument or being overridden.

2. The `StopProcessingIf` processor is imported into the `URLDownloader` processor. An additional, optional boolean input is added to the `URLDownloader` processor, named `STOP_IF_DOWNLOAD_UNCHANGED`. This means that users could add an additional input of `STOP_IF_DOWNLOAD_UNCHANGED`=`True` in their child recipes or overrides in order to stop processing after `URLDownloader` if the download is unchanged. The default behaviour is of course not altered.

Together, these two steps give AutoPkg users powerful additional flexibility to stop processes at the optimal time (i.e. no code reverification, repackaging or upload to repositories when not required), without changing all existing Download recipes to add a `StopProcessingIf` processor.

### DIFF
--- a/Code/autopkglib/StopProcessingIf.py
+++ b/Code/autopkglib/StopProcessingIf.py
@@ -33,11 +33,12 @@ class StopProcessingIf(Processor):
     description = __doc__
     input_variables = {
         "predicate": {
-            "required": True,
+            "required": False,
             "description":
                 ("NSPredicate-style comparison against an environment key. See "
                  "http://developer.apple.com/library/mac/#documentation/"
                  "Cocoa/Conceptual/Predicates/Articles/pSyntax.html"),
+            "default": 'FALSEPREDICATE'
         },
     }
     output_variables = {
@@ -56,7 +57,9 @@ class StopProcessingIf(Processor):
                 % (predicate_string, err))
 
         result = predicate.evaluateWithObject_(self.env)
-        self.output("(%s) is %s" % (predicate_string, result))
+        # Do not output when FALSEPREDICATE set
+        if predicate_string != 'FALSEPREDICATE':
+            self.output("(%s) is %s" % (predicate_string, result))
         return result
 
     def main(self):
--- a/Code/autopkglib/URLDownloader.py
+++ b/Code/autopkglib/URLDownloader.py
@@ -22,6 +22,7 @@ import time
 import xattr
 import tempfile
 
+from autopkglib.StopProcessingIf import StopProcessingIf
 from autopkglib import Processor, ProcessorError
 try:
     from autopkglib import BUNDLE_ID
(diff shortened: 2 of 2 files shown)


gregneagle:
I think this is not the right place for this. I understand your motivation, but the URLDownloader is a general purpose processor. I can imagine (and I think there might exist) recipes that might use URLDownloader multiple times. Please convince me.

grahampugh:
You are right that some recipes might use more than one URLDownloader processor, but it is an optional key with a default that would not change anything. For the vast majority of Download recipes that have just a single URLDownloader processor, we gain the additional functionality without anyone having to rewrite the recipes. 

For the recipes with more than one URLDownloader processor, if anyone wishes to benefit from the ability to stop the process, they could ask the author kindly to add the ability to override each value with a unique input, which is perfectly feasible, or else the requestor could clone the recipe into their own recipe to add the required keys, just like they currently have to do with ALL Download recipes at present to gain this flexibility.

If the process was instead moved to the beginning of the CodeSignatureVerification processor, as you suggest elsewhere, this would also be the case, since one would expect each downloaded file to be verified.  

Equally, if the option was added to the PkgCreator processor, the same if not further complexity would exist, as this processor could potentially have multiple inputs that could be the result of multiple URLDownload steps, or there could be multiple PkgCreator processors in some complicated circumstances. 

By the way, I've also developed a method of replicating the behaviour of MunkiImporter in JSSImporter, once I figure out which branch is "current", which would at least address the issues Jamf users are having with rewritten policies at every AutoPkg run. This PR would then be more about making large AutoPkg runs more efficient for those that wished to use such a workflow. 

I do understand that this change would be annoying since many recipe overrides would have to be reverified as a result of the code change, so I recognise that a lot of people are going to recognise the  benefit before this change would be accepted :)

MagerValp:
After reviewing the PR I don't think that this is the correct way to address the problem. I would much rather see improvements to the underlying issues with verification, packaging, and jss importing.

Am I understanding [your comment here](https://github.com/autopkg/recipes/issues/286#issuecomment-471310996) correctly in that the main issue of duplicate jss imports has been worked around? If so that's great, since it seemed like the hairiest one of the three.

Signature verification seems amenable to caching, either in an xattr or a small db or plist. A packaging cache would be more complex, but if it's really a pain point maybe it could be worth looking at?

PR #469, opened 2019-05-09
CURL/URLDownloader with flake8 + isort

nmcspadden:
I did both CURLDownloader and URLDownloader together because CURLDownloader just calls URLDownloader anyway.

### DIFF
--- a/Code/autopkglib/URLDownloader.py
+++ b/Code/autopkglib/URLDownloader.py
@@ -162,7 +162,7 @@ class URLDownloader(Processor):
         # this can cause issues if this item is eventually copied to a Munki repo
         # with the same permissions and the file is inaccessible by (for example)
         # the webserver.
-        os.chmod(pathname_temporary, "0644")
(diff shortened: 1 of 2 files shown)


Code/autopkglib/URLDownloader.py
```diff
@@ -162,7 +162,7 @@ def main(self):
         # this can cause issues if this item is eventually copied to a Munki repo
         # with the same permissions and the file is inaccessible by (for example)
         # the webserver.
-        os.chmod(pathname_temporary, 0644)
+        os.chmod(pathname_temporary, "0644")
```
gregneagle: This looks wrong.

gregneagle: `0o644` might be a better change if the goal is Python3-compatibility.

nmcspadden: You're totally right, let's use the right octal.

PR #470, opened 2019-05-09
CURL/URLTextSearcher with flake8 + isort

nmcspadden:


Code/autopkglib/CURLTextSearcher.py
```diff
@@ -18,11 +18,9 @@
...
 from autopkglib.URLTextSearcher import URLTextSearcher
 
-
 __all__ = ["CURLTextSearcher"]
 
 CURLTextSearcher = URLTextSearcher
 
 if __name__ == "__main__":
     PROCESSOR = CURLTextSearcher()
-    PROCESSOR.execute_shell()
```
gregneagle: So it deletes the line because it didn't end with a linefeed?

nmcspadden: Whoa!

PR #474, opened 2019-05-09
FileFinder with flake8 + isort

nmcspadden:


Code/autopkglib/FileFinder.py
```diff
@@ -16,13 +16,15 @@
 """See docstring for FileFinder class"""
 
 
+import os.path
 from glob import glob
-from autopkglib import Processor, ProcessorError
+
+from autopkglib import ProcessorError
```
gregneagle: Why did it drop "Processor" from the import?

nmcspadden: Because it's unused anywhere in the recipe. FileFinder inherits from DmgMounter instead.

gregneagle: OK. The GitHub code review tools make that difficult to investigate.

PR #486, opened 2019-05-09
MunkiInstallsItemCreator with flake8 + isort

nmcspadden:
Also replaced 'basestring' with 'str', as basestring no longer exists in py3. I verified that this still works in python2.

### DIFF
--- a/Code/autopkglib/MunkiInstallsItemsCreator.py
+++ b/Code/autopkglib/MunkiInstallsItemsCreator.py
@@ -105,7 +105,7 @@ class MunkiInstallsItemsCreator(Processor):
             for item in installs_array:
                 cmp_key = None
                 # If it's a string, set it for all installs items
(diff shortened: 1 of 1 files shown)


Code/autopkglib/MunkiInstallsItemsCreator.py
```diff
@@ -100,7 +105,7 @@ def create_installs_items(self):
             for item in installs_array:
                 cmp_key = None
                 # If it's a string, set it for all installs items
-                if isinstance(self.env["version_comparison_key"], basestring):
+                if isinstance(self.env["version_comparison_key"], str):
```
gregneagle: As in the other PR, this changes behavior in Python 2.

nmcspadden: I'm going to punt this down the line until better testing can be developed.

PR #489, opened 2019-05-09
PathDeleter with flake8+isort

nmcspadden:
Also replaced 'basestring' with 'str', as basestring no longer exists in py3. I verified that this still works in python2.

### DIFF
--- a/Code/autopkglib/PathDeleter.py
+++ b/Code/autopkglib/PathDeleter.py
@@ -40,7 +40,7 @@ class PathDeleter(Processor):
     def main(self):
         # if recipe writer gave us a single string instead of a list of strings,
         # convert it to a list of strings
-        if isinstance(self.env["path_list"], str):
+        if isinstance(self.env["path_list"], basestring):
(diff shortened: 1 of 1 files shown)


Code/autopkglib/PathDeleter.py
```diff
@@ -40,7 +40,7 @@ class PathDeleter(Processor):
     def main(self):
         # if recipe writer gave us a single string instead of a list of strings,
         # convert it to a list of strings
-        if isinstance(self.env["path_list"], basestring):
+        if isinstance(self.env["path_list"], str):
             self.env["path_list"] = [self.env["path_list"]]
```
gregneagle: No, this changes the behavior in Python 2.
`isiinstance(foo, basestring` is true if foo is a str or if it is unicode.

Your change would be false if self.env["path_list"] is unicode.

nmcspadden: I'm going to punt the issue down the line for now. I've switched it back to `basestring` until some better testing can be developed.

PR #498, opened 2019-05-09
SparkleUpdateInfoProvider with flake8+isort

nmcspadden:


Code/autopkglib/SparkleUpdateInfoProvider.py
```diff
@@ -233,7 +233,9 @@ def main(self):
         """Get URL for latest version in update feed"""
         def compare_version(this, that):
             """Compare loose versions"""
-            return cmp(LooseVersion(this), LooseVersion(that))
```
gregneagle: I think I know _why_ this code change is here, but I think there should be a comment. I don't find this terribly obvious or readable at first glance.

PR #514, opened 2019-05-10
generate_processor_docs with flake8+isort

nmcspadden:


nmcspadden:
whoa wait this isn't the right file

Issue #518, opened 2019-05-11
Summary result data doesn't support multiple items


gkluoe:
I have a custom processor which creates many packages at once - I'd like to be able to list them in AutoPkg's summary output. As far as I can tell, though, it's not possible to have anything other than a single `dict` as the data item in a summary report from a single processor:

https://github.com/autopkg/autopkg/blob/d159a8e0b747e6ffe79300457dce3ccde39c3d73/Code/autopkg#L2129-L2140

It looks to me as if the intention is that multiple runs of the same processor could add to `data_rows`, but there's no way for a single processor to add multiple rows (eg because it produced several packages it wants to tell you about).

I'd like to propose supporting a list of `dicts` as the data for a summary result, with each `dict` being added as a `data_row`.

gregneagle:
You are right in your analysis: to-date we've never had a processor that produced multiple things of interest. Even an existing recipe that imports multiple packages to a Munki repo (https://github.com/autopkg/recipes/blob/master/munkitools/munkitools3.munki.recipe) calls the MunkiImporter processor five times, which can add up to five items to the summary results.

gkluoe:
Hmm, thanks. Having thought about this more, I suspect that what I'm doing (one processor creating lots of packages) is an anti-pattern, and probably I should re-engineer it to produce only a single package at a time. That's simple enough.

The processor is for Native Instruments software, where the suites tend to have large numbers of individual products - 54 for the suite I'm dealing with (and that's which is far from the largest).

Perhaps I should include a script that folk can use to easily generate individual recipes for each product in the suite they want to package - that way the suite can be packaged in one AutoPkg run; the current reporting mechanism works fine; and the individual recipes can be overridden if the user wants to process the packages further.

https://github.com/UoE-macOS/autopkg-recipes/tree/master/Native%20Instruments

nmcspadden:
This issue doesn't seem to have any action items left. I'm closing it, but please re-open if there's more that we need to do. I agree with your general assessment that AutoPkg isn't really suited for handling a single processor producing multiple things of interest, and recipes that produce multiple output is a bit tough to fit into the current architecture.

PR #519, opened 2019-05-11
Support a list of dicts in summary output (#518)

gkluoe:
As per issue #518 here's a possible implementation that would (I think!) leave current behaviour unchanged for current recipes but would allow a processor to provide a list of dicts of data for the summary table.

### DIFF
--- a/Code/autopkg
+++ b/Code/autopkg
@@ -2133,11 +2133,15 @@ def run_recipes(argv):
                     if key not in summary_results:
                         summary_results[key] = {}
                         summary_results[key]['summary_text'] = summary_text
-                        if type(data).__name__ in ['dict', '__NSCFDictionary']:
+                        if type(data).__name__ != 'list':
+                            data = [data]
+                        if type(data[0]).__name__ in ['dict', '__NSCFDictionary']:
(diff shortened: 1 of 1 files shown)


nmcspadden:
Can you demonstrate the results of this? 

Some suggestions for testing this adequately:
- Please provide a sample of input summary_results (or the code of your custom processor that creates multiple packages)
- Show what happens when run on master (i.e. what the undesirable behavior is that happens right now)
- Show what happens when run with your code
- Show the results when running a recipe that doesn't include your custom processor (i.e. any core recipe in the public repo), to prove that it doesn't change existing behavior

gkluoe:
Thanks for the helpful reply! After thinking more and consulting on Slack, I've come up with what I think is a more idiomatic way of handling suites of dozens of packages:

https://github.com/UoE-macOS/autopkg-recipes/blob/master/Native%20Instruments/README.md

I'm happy for this PR to be closed, though if it's actually thought to be useful I'm happy to develop some tests as suggested.

gregneagle:
Since you no longer need this, and no-one else yet does, we'll close this.

PR #523, opened 2019-05-16
Moving sidebar into separate function

nmcspadden:
The purpose of this is to separate out some of the existing functionality into smaller functions, for added testability in the future. The primary and more immediate objective is to allow taking different actions based on the results. The next PR in this stack will be adding an argument to select a specific processor to update.

Testing by running:
```
./generate_processor_docs.py 1.0.5
Cloned to /var/folders/th/3tqg05g54f73x0k_mstm5wy1_0g_xw/T/tmp7bf6sD.


commit df99d8a005bd4350b8662f4e794c1df53d10d0fa
Author: Nick McSpadden <nmcspadden@gmail.com>
Date:   Thu May 16 14:10:54 2019 -0700

    Updating Wiki docs for release 1.0.5

diff --git a/Processor-AppPkgCreator.md b/Processor-AppPkgCreator.md
index f2a1a4f..09abb19 100644
--- a/Processor-AppPkgCreator.md
+++ b/Processor-AppPkgCreator.md
@@ -12,7 +12,7 @@ Calls autopkgserver to create a package from an application.
     - **description:** Path to an application to be packaged. Can be on a disk image and globbed. If not set, defaults to %pathname%/\*.app. Typically %pathname% points to a disk image downloaded in a prior recipe step. - **version:**
     - **required:** False
-    - **description:** Version of the app. If not set, will be extracted from theCFBundleShortVersionString in the app's Info.plist.
+    - **description:** Version of the app. If not set, will be extracted from the CFBundleShortVersionString in the app's Info.plist.
 - **force\_pkg\_build:**
     - **required:** False
     - **description:** When set, this forces building a new package even if a package already exists in the output directory with the same identifier and version number. Defaults to False
diff --git a/Processor-InstallFromDMG.md b/Processor-InstallFromDMG.md
index de2100f..e3a9532 100644
--- a/Processor-InstallFromDMG.md
+++ b/Processor-InstallFromDMG.md
@@ -7,7 +7,7 @@ Calls autopkginstalld to copy items from a disk image to the root
 ## Input Variables
 - **items\_to\_copy:**
     - **required:** True
-    - **description:** Array of dictionaries describing what is to be copied. Each item should contain 'source\_item' and 'destination\_path', and may optionally include: 'destination\_item' to rename the item on copy, and 'user', 'group' and 'mode' to explictly set those items.
+    - **description:** Array of dictionaries describing what is to be copied. Each item should contain 'source\_item' and 'destination\_path', and may optionally include: 'destination\_item' to rename the item on copy, and 'user', 'group' and 'mode' to explicitly set those items.
 - **dmg\_path:**
     - **required:** True
     - **description:** Path to the disk image.
...

-------------------------------------------------------------------

Shown above is the commit log for the changes to the wiki markdown.
Type 'push' to accept and push the changes to GitHub. The wiki repo
local clone can be also inspected at:
/var/folders/th/3tqg05g54f73x0k_mstm5wy1_0g_xw/T/tmpW_spSy.
```

### DIFF
--- a/Scripts/generate_processor_docs.py
+++ b/Scripts/generate_processor_docs.py
@@ -97,6 +97,43 @@ def indent_length(line_str):
     return len(line_str) - len(line_str.lstrip())
 
 
+def generate_sidebar(sidebar_path):
+    """Generate new _Sidebar.md contents."""
+    # Generate the Processors section of the Sidebar
+    processor_heading = "  * **Processor Reference**"
+    toc_string = ""
+    toc_string += processor_heading + "\n"
+    for processor_name in sorted(processor_names(), key=lambda s: s.lower()):
+        page_name = "Processor-%s" % processor_name
+        page_name.replace(" ", "-")
+        toc_string += "      * [[%s|%s]]\n" % (processor_name, page_name)
+
+    with open(sidebar_path, "r") as fdesc:
+        current_sidebar_lines = fdesc.read().splitlines()
+
+    # Determine our indent amount
+    section_indent = indent_length(processor_heading)
+
+    past_processors_section = False
+    for index, line in enumerate(current_sidebar_lines):
+        if line == processor_heading:
+            past_processors_section = True
+            processors_start = index
+        if (
+            (indent_length(line) <= section_indent)
+            and past_processors_section
+        ):
+            processors_end = index
+
+    # Build the new sidebar
+    new_sidebar = ""
+    new_sidebar += "\n".join(current_sidebar_lines[0:processors_start]) + "\n"
+    new_sidebar += toc_string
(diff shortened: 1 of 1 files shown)


gregneagle:
Did you redo your test after updating the code? The sorting doesn't appear to indicate that you did (neither does the date!)

nmcspadden:
I'm bad at copy-pasting. And yes, I did, but the wiki sidebar is currently sorted so it doesn't think there are any changes to make.

nmcspadden:
I updated the test with a newer paste.

gregneagle:
"wiki sidebar is currently sorted so it doesn't think there are any changes to make." and that's a good test!

PR #526, opened 2019-05-17
Blackening all Python code

nmcspadden:
The time has come. This runs `black` through all code in AutoPkg.

Since there are no decisions to make, and no choices about what we like, don't like, etc., it's pretty much an all-or-nothing. I don't think there's that much value in doing this file-by-file, because there will be no difference in the outcome.

This was accomplished by installing Python3 using Greg Neagle's relocatable-python framework and using pip to install black, isort, and flake8, and then using the black from Python 3 to run on the autopkg github checkout.

### DIFF
--- a/Code/FoundationPlist/FoundationPlist.py
+++ b/Code/FoundationPlist/FoundationPlist.py
@@ -43,43 +43,55 @@ and writePlistToString().
 
 import os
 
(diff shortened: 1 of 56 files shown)


gregneagle:
YOLO

gregneagle:
BTW, this formatting is likely to drift over time in at least the short term since we have no way to enforce it and only somewhat vague directions on how others can achieve the same result.

PR #527, opened 2019-05-20
Log instead of print

MichalMMac:
This PR is the implementation of [suggestion](https://groups.google.com/d/msg/autopkg-discuss/YcO2oQR5JME/-19T870MAgAJ) in Google group [thread](https://groups.google.com/forum/#!topic/autopkg-discuss/YcO2oQR5JME) about Python2 EOL..

It does two things:

- ~~Moves `log()` and `log_err()` functions from `autopkg` to `autopkglib`. ba270f1~~ Already in master
- Replaces `print()` function with `log()` and `log_err()` where it makes sense. 1af50fd6d24ba8179227c053164ebef3c07c8ac8
- Remove now unnecessary `print_function` imports da82c92626d881415d81a15b76854633852a41fd
- Switch `ID NOT FOUND` message from stdout to stderr 5e06c381ed8b0bbfa3a6026566fd89715d824e93

### DIFF
--- a/Code/autopkglib/AppDmgVersioner.py
+++ b/Code/autopkglib/AppDmgVersioner.py
@@ -15,8 +15,6 @@
 # limitations under the License.
 """See docstring for AppDmgVersioner class"""
 
-from __future__ import print_function
-
 import glob
 import os.path
 
--- a/Code/autopkglib/AppDmgVersioner.py
+++ b/Code/autopkglib/AppDmgVersioner.py
@@ -20,20 +20,19 @@ from __future__ import print_function
 import glob
 import os.path
 
-from autopkglib import ProcessorError
+from autopkglib import ProcessorError, log
 from autopkglib.DmgMounter import DmgMounter
-
 
 # pylint: disable=no-name-in-module
 try:
     from Foundation import NSData, NSPropertyListSerialization
     from Foundation import NSPropertyListMutableContainers
 except:
-    print(
+    log(
         "WARNING: Failed 'from Foundation import NSData, "
         "NSPropertyListSerialization' in " + __name__
     )
-    print(
+    log(
         "WARNING: Failed 'from Foundation import "
         "NSPropertyListMutableContainers' in " + __name__
     )
--- a/Code/autopkglib/MunkiSetDefaultCatalog.py
+++ b/Code/autopkglib/MunkiSetDefaultCatalog.py
@@ -16,14 +16,15 @@
 # limitations under the License.
 """See docstring for MunkiSetDefaultCatalog class"""
 
-from autopkglib import Processor
+from __future__ import print_function
 
+from autopkglib import Processor, log
 
 # pylint: disable=no-name-in-module
 try:
     from Foundation import CFPreferencesCopyAppValue
 except:
-    print(
+    log(
         "WARNING: Failed 'from Foundation import CFPreferencesCopyAppValue' "
         "in " + __name__
     )
--- a/Code/autopkglib/MunkiSetDefaultCatalog.py
+++ b/Code/autopkglib/MunkiSetDefaultCatalog.py
@@ -16,8 +16,6 @@
 # limitations under the License.
 """See docstring for MunkiSetDefaultCatalog class"""
 
-from __future__ import print_function
-
 from autopkglib import Processor, log
 
 # pylint: disable=no-name-in-module
--- a/Code/autopkglib/__init__.py
+++ b/Code/autopkglib/__init__.py
@@ -78,10 +78,8 @@ try:
         kCFPreferencesCurrentHost,
     )
 except:
-    print(
-        "WARNING: Failed 'from Foundation import NSArray, NSDictionary' in " + __name__
-    )
-    print(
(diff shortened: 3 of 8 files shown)


Code/autopkglib/MunkiSetDefaultCatalog.py
```diff
@@ -16,14 +16,16 @@
 # limitations under the License.
 """See docstring for MunkiSetDefaultCatalog class"""
 
-from autopkglib import Processor
+from __future__ import print_function
```
nmcspadden: This isn't actually necessary anymore if you're replacing print().

nmcspadden: This isn't necessary anymore if you're replacing print() with log().

Code/autopkglib/__init__.py
```diff
@@ -420,7 +429,7 @@ def get_recipe_identifier(self, recipe):
         """Return the identifier given an input recipe plist."""
         identifier = recipe.get("Identifier") or recipe["Input"].get("IDENTIFIER")
         if not identifier:
-            print("ID NOT FOUND")
+            log("ID NOT FOUND")
```
nmcspadden: Although this was originally to stdout, this seems like it should logically go to stderr.

Code/autopkglib/MunkiSetDefaultCatalog.py
MichalMMac: Ok. So just to verify. Processor imports the `log()` from `autopkglib`. `log()` uses `print()` but we don't need to import the `print()` because it was already imported in `autopkg` [itself](https://github.com/autopkg/autopkg/blob/eaf5ae258ab921ca6dec21e1eddb674e920825ea/Code/autopkg#L19). Correct?

I also removed this import from others processors. See f584460.

Code/autopkglib/__init__.py
MichalMMac: Ok. 988df72

MichalMMac:
> Please rebase this PR and ensure it passes the new black + isort requirements.

I set up the evironment as described in [CONTRIBUTING.md] (https://github.com/autopkg/autopkg/blob/master/CONTRIBUTING.md)
I fixed all black and isort problems and relevant flake8 problems. See 986113d c59ae4d

But there are some unresolved problems flake8 detected. For example F821 about `basestring` and `unicode`. These are not relevant to my changes so I had to commit with `--no-verify` switch.

nmcspadden:
Let's back up a bit. This PR is now way too huge and involves way too many changes that are out of scope for the actual problem. 

If the goal of this PR is to swap the print() statements for log(), please submit one that only includes those changes. The rest of the flake8 errors and isort issues throughout the rest of the codebase should be handled in separate PRs. 

In general, it's better to make multiple small PRs for smaller and singular changes than to make giant PRs that cover lots of ground. It's much more difficult to review and test larger sweeping changes.

nmcspadden:
https://github.com/autopkg/autopkg/pull/535 should fix the issue where flake8 warns on all the Python3-specific errors. I've now limited to py2 only.

MichalMMac:
Moved all those isort newlines "fixes" to #536.
This PR should be ready to merge.

Code/autopkglib/__init__.py
gregneagle: I think the point is more that since you are no longer calling `print()`, having replaced all those calls with `log()`, you no longer need to `from __future__ import print_function`, and if you leave that in, the various linters might complain about unused imports.

Code/autopkglib/AppDmgVersioner.py
```diff
@@ -15,25 +15,22 @@
...
 from autopkglib.DmgMounter import DmgMounter
 
-
 # pylint: disable=no-name-in-module
 try:
     from Foundation import NSData, NSPropertyListSerialization
     from Foundation import NSPropertyListMutableContainers
 except:
-    print(
+    log(
```
nmcspadden: All processors should use self.output() instead of log/log_err.

Code/autopkglib/MunkiSetDefaultCatalog.py
gregneagle: I think the point is more that since you are no longer calling print(), having replaced all those calls with log(), you no longer need to from __future__ import print_function, and if you leave that in, the various linters might complain about unused imports.

Code/autopkglib/AppDmgVersioner.py
gregneagle: On the surface that sounds correct/reasonable. There _might_ be reasons I have forgotten where a processor needs to bypass self.output(), however. So I'm hesitant to say: "sure, go ahead and do a mass find-and-replace". 

Let's not get so enamored of our consistency that we introduce regressions because we quickly "conform" all the code.

nmcspadden: That sounds like a reasonable response to me. We can always come by later and change individual processors if there's a feeling that they're too "noisy" or not respectful enough of AutoPkg verbosity settings.

nmcspadden:
I'm totally okay with this as-is, thanks for the work.

nmcspadden:
@gregneagle seem reasonable to you?

PR #529, opened 2019-05-31
Fixing flake8 errors

nmcspadden:
Moving log and log_err to autopkglib so flake8 doesn't complain anymore.

### DIFF
--- a/Code/autopkg
+++ b/Code/autopkg
@@ -47,11 +47,12 @@ from autopkglib import (
     get_identifier,
     get_pref,
     get_processor,
+    log,
+    log_err,
     processor_names,
     set_pref,
     version_equal_or_greater,
 )
-
 
 if sys.platform != "darwin":
     print(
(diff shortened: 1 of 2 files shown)


Code/autopkg
```diff
@@ -2365,7 +2353,9 @@ def run_recipes(argv):
                 # record any summary results
                 output_keys = item["Output"].keys()
                 results_keys = [
-                    key for key in output_keys if key.endswith("_summary_result")
+                    summary_key
+                    for summary_key in output_keys
+                    if key.endswith("_summary_result")
```
homebysix: Should this not be `summary_key` also?

gregneagle: Yep, looks like it to me.

gregneagle: I thought Nick was just reformatting and moving functions to other modules. Sigh

homebysix: Resolved by https://github.com/autopkg/autopkg/pull/531/files

PR #530, opened 2019-05-31
Adding flake8 to pre-commit config

nmcspadden:


gregneagle:
Let's fix the breakage from the earlier commit first...

PR #533, opened 2019-05-31
Adding platform-specific determination functions

nmcspadden:
This opens the door for AutoPkg on alternative platforms. The memoization is so that we don't have to make another call to platforms() every time we want to do this lookup - the result is cached in each invocation.

### DIFF
--- a/Code/autopkglib/__init__.py
+++ b/Code/autopkglib/__init__.py
@@ -21,6 +21,7 @@ from __future__ import print_function
 import glob
 import imp
 import os
+import platform
 import pprint
 import re
 import subprocess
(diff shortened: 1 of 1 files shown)


Code/autopkglib/__init__.py
```diff
@@ -29,6 +30,43 @@
 
 import FoundationPlist
 
+
+def memoize(func):
+    """
+    Decorator to memoize results of function calls.
+    Works with functions that have no args.
+    https://www.thumbtack.com/engineering/a-primer-on-python-decorators/
+    """
```
gregneagle: why not use an implementation that works with functions that _have_ args, too?

https://github.com/munki/munki/blob/master/code/client/munkilib/utils.py#L33-L44

nmcspadden: Awesome, I'll use that instead.

PR #534, opened 2019-05-31
Allowing URLDownloader to be imported on non-Mac platforms

nmcspadden:
xattr module doesn't exist on Windows, so this gates it to a Mac-only import. There's no current expectation that any of this will run on other platforms yet. This is just to allow importing of AutoPkg processors to work on alternate platforms without error (so unit tests can be run on non-macOS, to start with).

### DIFF
--- a/Code/autopkglib/URLDownloader.py
+++ b/Code/autopkglib/URLDownloader.py
@@ -20,14 +20,10 @@ import subprocess
 import tempfile
 import time
 
-import xattr
-from autopkglib import Processor, ProcessorError
+from autopkglib import BUNDLE_ID, Processor, ProcessorError, is_mac
 
-
-try:
-    from autopkglib import BUNDLE_ID
-except ImportError:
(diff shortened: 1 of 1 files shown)


nmcspadden:
Does this look good?

gregneagle:
I assume eventually you plan to add non-Mac functionality to this processor, otherwise it could just be

```python
try:
    import xattr
except ImportError:
    raise ProcessorError("This processor only runs on a Mac!")
```

and no need for the `is_mac()` function at all

nmcspadden:
That is the long-term goal, yes. As-is, curl on Windows is a stub to PowerShell's Invoke-WebRequest and it doesn't work as a drop-in substitute.

gregneagle:
Pretty sure you can get a "full" curl on Windows or I'd think no-one would be running reposado on Windows (and I know it's been/being done).

nmcspadden:
Yes you can, my win branch on my fork requires you to install curl separately, but I'm not yet settled on that solution.

PR #538, opened 2019-06-14
Adding some basic unit tests to AutoPkg

nmcspadden:
This gets the template in place for doing unit testing for AutoPkg. To start with, I provided a unit test for one Processor (FileFinder, because it's easy)

To run the tests, you can do it using the unittest module via the command line. You can do it in a virtualenv, or if you feel like messing with system python, you can install mock there:
```
$ PATH=/usr/bin python -m unittest discover
..............
----------------------------------------------------------------------
Ran 14 tests in 0.026s

OK
```

If you want verbose results for the tests:
```
$ PATH=/usr/bin python -m unittest discover -v
test_get_identifier_from_recipe_file_returns_identifier (tests.test_autopkglib.TestAutoPkg)
get_identifier_from_recipe-file should return identifier. ... ok
test_get_identifier_from_recipe_file_returns_none (tests.test_autopkglib.TestAutoPkg)
get_identifier_from_recipe-file should return None if no identifier. ... ok
test_get_identifier_returns_identifier (tests.test_autopkglib.TestAutoPkg)
get_identifier should return the identifier. ... ok
test_get_identifier_returns_none (tests.test_autopkglib.TestAutoPkg)
get_identifier should return None if no identifier is found. ... ok
test_is_linux_returns_false_on_not_linux (tests.test_autopkglib.TestAutoPkg)
On not-Linux, is_linux() should return False. ... ok
test_is_linux_returns_true_on_linux (tests.test_autopkglib.TestAutoPkg)
On Linux, is_linux() should return True. ... ok
test_is_mac_returns_false_on_not_mac (tests.test_autopkglib.TestAutoPkg)
On not-macOS, is_mac() should return False. ... ok
test_is_mac_returns_true_on_mac (tests.test_autopkglib.TestAutoPkg)
On macOS, is_mac() should return True. ... ok
test_is_windows_returns_false_on_not_windows (tests.test_autopkglib.TestAutoPkg)
On not-Windows, is_windows() should return False. ... ok
test_is_windows_returns_true_on_windows (tests.test_autopkglib.TestAutoPkg)
On Windows, is_windows() should return True. ... ok
test_found_a_dmg_match (tests.test_filefinder.TestFileFinder)
If we find a match inside a DMG, it should be in the env. ... ok
test_found_a_match (tests.test_filefinder.TestFileFinder)
If we find a match, it should be in the env. ... ok
test_no_fail_if_good_env (tests.test_filefinder.TestFileFinder)
test_raise_if_not_glob (tests.test_filefinder.TestFileFinder)
Raise an exception if glob is not passed to find_method. ... ok

----------------------------------------------------------------------
Ran 14 tests in 0.031s

OK
```

A screenshot from the test runner in Visual Studio Code:
tests

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,8 +1,9 @@
 # .DS_Store files!
 .DS_Store
 
-# don't track .pyc files
+# don't track .pyc files or autopkgc
 *.pyc
+autopkgc
 
 # IDE-specific files
 .vscode
--- a/Code/autopkglib/__init__.py
+++ b/Code/autopkglib/__init__.py
@@ -78,8 +78,10 @@ try:
         kCFPreferencesCurrentHost,
     )
 except:
-    log("WARNING: Failed 'from Foundation import NSArray, NSDictionary' in " + __name__)
-    log(
+    print(
+        "WARNING: Failed 'from Foundation import NSArray, NSDictionary' in " + __name__
+    )
+    print(
         "WARNING: Failed 'from CoreFoundation import "
         "CFPreferencesAppSynchronize, ...' in " + __name__
     )
--- a/Code/tests/test_autopkglib.py
+++ b/Code/tests/test_autopkglib.py
@@ -1,0 +1,208 @@
+#!/usr/bin/env python
+
+import imp
+import plistlib
+import unittest
+from textwrap import dedent
+
+from mock import patch
+
+
+# DO NOT MOVE THIS! This needs to happen BEFORE importing autopkglib
+patch("autopkglib.memoize", lambda x: x).start()
+import autopkglib  # isort: skip
+
+autopkg = imp.load_source("autopkg", "autopkg")
+
+
+class TestAutoPkg(unittest.TestCase):
+    """Test class for AutoPkglib itself."""
+
+    def setUp(self):
+        # This forces autopkglib to accept our patching of memoize
+        imp.reload(autopkglib)
+        self.download_recipe = dedent(
+            """\
+            <?xml version="1.0" encoding="UTF-8"?>
+            <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+            <plist version="1.0">
+            <dict>
+                <key>Description</key>
+                <string>Downloads latest Google Chrome disk image.</string>
+                <key>Identifier</key>
+                <string>com.github.autopkg.download.googlechrome</string>
+                <key>Input</key>
+                <dict>
+                    <key>NAME</key>
+                    <string>GoogleChrome</string>
+                    <key>DOWNLOAD_URL</key>
+                    <string>https://dl.google.com/chrome/mac/stable/GGRO/googlechrome.dmg</string>
+                </dict>
+                <key>MinimumVersion</key>
+                <string>0.2.0</string>
+                <key>Process</key>
+                <array>
+                    <dict>
+                        <key>Processor</key>
+                        <string>URLDownloader</string>
+                        <key>Arguments</key>
+                        <dict>
+                            <key>url</key>
+                            <string>%DOWNLOAD_URL%</string>
+                            <key>filename</key>
+                            <string>%NAME%.dmg</string>
+                        </dict>
+                    </dict>
+                    <dict>
+                        <key>Processor</key>
+                        <string>EndOfCheckPhase</string>
+                    </dict>
+                    <dict>
+                        <key>Processor</key>
+                        <string>CodeSignatureVerifier</string>
+                        <key>Arguments</key>
+                        <dict>
+                            <key>input_path</key>
+                            <string>%pathname%/Google Chrome.app</string>
+                            <key>strict_verification</key>
+                            <false/>
+                            <key>requirement</key>
+                            <string>(identifier "com.google.Chrome" or identifier "com.google.Chrome.beta" or identifier "com.google.Chrome.dev" or identifier "com.google.Chrome.canary") and (certificate leaf = H"85cee8254216185620ddc8851c7a9fc4dfe120ef" or certificate leaf = H"c9a99324ca3fcb23dbcc36bd5fd4f9753305130a")</string>
+                        </dict>
+                    </dict>
+                </array>
+            </dict>
+            </plist>
+        """
+        )
+        self.munki_recipe = dedent(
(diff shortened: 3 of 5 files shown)


.gitignore
```diff
@@ -1,8 +1,9 @@
 # .DS_Store files!
 .DS_Store
 
-# don't track .pyc files
+# don't track .pyc files or autopkgc
 *.pyc
+autopkgc
```
nmcspadden: This is autogenerated when importing autopkg directly via `imp.load_source()` for unit testing.

Code/autopkglib/__init__.py
```diff
@@ -78,8 +78,10 @@ def is_linux():
         kCFPreferencesCurrentHost,
     )
 except:
-    log("WARNING: Failed 'from Foundation import NSArray, NSDictionary' in " + __name__)
-    log(
+    print(
```
nmcspadden: These were converted back to print() because log() isn't defined yet when this import runs.

Code/tests/test_autopkglib.py
```diff
@@ -0,0 +1,208 @@
...
+import imp
+import plistlib
+import unittest
+from textwrap import dedent
+
+from mock import patch
+
+
+# DO NOT MOVE THIS! This needs to happen BEFORE importing autopkglib
+patch("autopkglib.memoize", lambda x: x).start()
```
nmcspadden: It took me a while to figure this out. Memoizing wraps the function at module import time, so the values for `is_mac()`, `is_windows()`, etc. were all predefined at import time. To do effective unit testing on a memoized function, we have to essentially apply a patch on memoize itself. By doing this here, we are overlaying a patch for `autopkglib.memoize` that always just returns what it's called with, rather than using the one defined in autopkglib.

This has to be defined prior to autopkglib being imported, so the patch is present on first load.

Code/tests/test_autopkglib.py
```diff
@@ -0,0 +1,208 @@
...
+import autopkglib  # isort: skip
+
+autopkg = imp.load_source("autopkg", "autopkg")
```
nmcspadden: We do this because autopkg itself is not a module, but can be treated like one.

Code/tests/test_autopkglib.py
```diff
@@ -0,0 +1,208 @@
...
+
+
+class TestAutoPkg(unittest.TestCase):
+    """Test class for AutoPkglib itself."""
+
+    def setUp(self):
+        # This forces autopkglib to accept our patching of memoize
+        imp.reload(autopkglib)
```
nmcspadden: With the patch created above on line 12, we can now force-reload autopkglib prior to each test being run to reset the memoized values. This allows us to test multiple different outcomes for a memoized function.

Code/tests/test_autopkglib.py
```diff
@@ -0,0 +1,208 @@
...
+                </array>
+            </dict>
+            </plist>
+        """
+        )
+
+    def tearDown(self):
+        pass
+
+    @patch("autopkglib.platform.platform")
```
nmcspadden: This only works now because `memoize` is now patched to return the function it's called with, rather than store the results and return the cache.

Code/tests/test_filefinder.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+class TestFileFinder(unittest.TestCase):
+    """Test class for FileFinder Processor."""
+
+    def setUp(self):
+        self.good_env = {"find_method": "glob", "pattern": "test"}
+        self.bad_env = {"find_method": "fake"}
+        self.input_plist = StringIO.StringIO(plistlib.writePlistToString(self.good_env))
+        self.processor = FileFinder(infile=self.input_plist)
```
nmcspadden: This is the general framework for testing a Processor - you add it as a class attribute in the setup, and then you can use it whenever you call main().

Code/tests/test_filefinder.py
```diff
@@ -0,0 +1,63 @@
...
+        self.input_plist.close()
+
+    def test_raise_if_not_glob(self):
+        """Raise an exception if glob is not passed to find_method."""
+        self.processor.env = self.bad_env
+        with self.assertRaises(ProcessorError):
+            self.processor.main()
+
+    @mock.patch("autopkglib.FileFinder.globfind")
+    def test_no_fail_if_good_env(self, mock_glob):
```
nmcspadden: I chose FileFinder to exemplify this because it's an easy example, but if there's a processor that actually makes changes to things on disk - such as something like PkgBuilder, FileMover, or Copier, etc. - all of those things have to be mocked so that the unit test itself doesn't actually do anything in production. 

So this is an easy example, but unfortunately not be a comprehensive enough example.

Code/tests/test_filefinder.py
```diff
@@ -0,0 +1,63 @@
...
+
+    @mock.patch("autopkglib.FileFinder.globfind")
+    def test_found_a_match(self, mock_glob):
+        """If we find a match, it should be in the env."""
+        self.processor.env = self.good_env
+        mock_glob.return_value = "test"
+        self.processor.main()
+        self.assertEqual(self.processor.env["found_filename"], "test")
+
+    @mock.patch("autopkglib.FileFinder.unmount")
```
nmcspadden: order to test.This is an example of patching things that actually change system state - like mounting and unmounting a DMG. We patch the mount and unmount functions here, and predefine their return values to provide a fixed path. That way, they don't depend on actual DMGs being present in order to test.

python2_requirements.txt
```diff
@@ -6,13 +6,15 @@ contextlib2==0.5.5
...
 flake8==3.7.7
+funcsigs==1.0.2
 functools32==3.2.3.post2
 futures==3.2.0
 identify==1.4.3
 importlib-metadata==0.13
 importlib-resources==1.0.2
 isort==4.3.20
 mccabe==0.6.1
+mock==3.0.5
```
nmcspadden: This is incorporated into the unittest module in python 3, but in python 2 is external.

nmcspadden:
@homebysix or @gregneagle - any objections or questions? I'm happy to add more commentary if more explanation is needed.

homebysix:
I'm going to take some time to wrap my head around the tests, then will probably try to write a few for other modules.

I added some very basic steps for setting up and running unit tests on the [wiki](https://github.com/autopkg/autopkg/wiki/Running-Unit-Tests).

PR #446, opened 2019-01-20
URLGetter refactoring and HTTP proxy fix

MichalMMac:
My original goal was to fix problem with autopkg not downloading files through HTTP proxy #416 
However when going through the code I found out it was somewhat unsatisfactory (very long functions, bad coding habits , etc.). Only thing to do was the *Right Thing* -> bit of refactoring.

- This PR introduces `URLGetter` class (suggested in https://github.com/autopkg/autopkg/issues/416#issuecomment-386764735). `URLGetter` is responsible for launching `curl` via subprocess and handling generic `curl` functionality like parsing HTTP headers. Classes `URLDownloader`, `URLTextSearcher`, `GitHubSession`, and `SparkleUpdateInfoProvider` are now subclasses of `URLGetter`. 
- HTTP header parsing is now done after the download. There is no infinite loop with break conditions, polling the subprocesses for output and magic constants like `maxheaders`.
- Code in classes  `URLDownloader`, `URLTextSearcher`, `GitHubSession`, and `SparkleUpdateInfoProvider` has been refactored. I tried to split it into sensible methods. Almost all `pylint` issues are now fixed.
- Refactoring the code handling `subprocess` and HTTP header parsing Fixes #416 by itself
- Fixes [issue](https://groups.google.com/forum/#!topic/autopkg-discuss/E6o9iHsSLSs) with redownloading files when using proxy 
- `curl_opts` input variable is now present in all subclasses of `URLGetter`.
- ~~https://github.com/MichalMMac/autopkg/commit/c0024278099cf3841a776c9005a7426ca0df9802  attempts to deal with #432. There is  a new `rename_after_redirect` input variable in `URLDownloader`. See the description.~~

I tried to [reach out](https://groups.google.com/forum/#!topic/autopkg-discuss/i4FswIat6GA) to AutoPkg google group and MacSysAdmin slack channel without much success. Fellow admins with same problem #416 tested my code successfully. However I belive this PR could use more testing (and RC version if merged).

Issue #416, opened 2018-05-03
Not working with HTTP proxy since 1.0.4


MichalMMac:
I use AutoPkg on macOS which always uses HTTP(S) proxy for HTTP(S) connections.
Proxy is configured via environment variables. Since 1.0.4 I get following crash when doing `autopkg searxh xyz` or when running some recipes which connect to Github. This is probably related to change `GitHub API requests are now performed using curl. ` made in 1.0.4.

```
Traceback (most recent call last):
  File "/usr/local/bin/autopkg", line 2453, in <module>
    sys.exit(main(sys.argv))
  File "/usr/local/bin/autopkg", line 2447, in main
    exit(subcommands[verb]['function'](argv))
  File "/usr/local/bin/autopkg", line 912, in search_recipes
    results = do_gh_code_search(query, use_token=options.use_token)
  File "/usr/local/bin/autopkg", line 826, in do_gh_code_search
    accept="application/vnd.github.v3.text-match+json")
  File "/Library/AutoPkg/autopkglib/github/__init__.py", line 209, in call_api
    resp_data = json.loads(page_content)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/json/__init__.py", line 338, in loads
    return _default_decoder.decode(s)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/json/decoder.py", line 366, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/json/decoder.py", line 384, in raw_decode
    raise ValueError("No JSON object could be decoded")
ValueError: No JSON object could be decoded
``` 

macOS: 10.13.4
Clean install with comand-line tools installed.
Current workaround: Downgrade to 1.0.3

Tugzrida:
Not working for me either. I did some investigating and the problem is in `autopkglib.github`. By switching to curl instead of urllib and manually parsing the headers, it appears to be getting confused by the headers returned when using a proxy. I added some verbose output to dump the entire raw data from the curl command this is the first bit of output:

    HTTP/1.1 200 Connection Established
    
    Proxy-Agent: Zscaler/6.0
    
    
    
    HTTP/1.1 200 OK
    
    Date: Fri, 04 May 2018 03:52:47 GMT
    ...

[...]

PR #544, opened 2019-07-04
URLGetter processor

MichalMMac:
This PR was split from #446 to simplify the review.

Goal of `URLGetter` processor is to implement common funcionality used by following items in one place:

- `GitHubReleasesInfoProvider` processor (See changes in #549)
- `GitHubSession` class (See change in #548)
- `SparkleUpdateInfoProvider` processor (See change in #547)
- `URLDownloader` processor (See change in #545)
- `URLTextSearcher` processor  (See change in #546 )
 
Basically we want to deduplicate similiar code handling following:

- Selecting desired `curl` binary to be used.
- Adding items to `curl_cmd` array like `--header` and other custom `curl` options.
- Using subprocess to launch `curl`
- Parse `curl` output including HTTP headers, FTP headers and possible errors

`URLGetter` processor is **not intended to be used directly** by recipes. Instead it should be
a superclass for other processors. This was suggested in https://github.com/autopkg/autopkg/issues/416#issuecomment-386764735.

### DIFF
--- a/Code/autopkglib/URLGetter.py
+++ b/Code/autopkglib/URLGetter.py
@@ -1,0 +1,191 @@
+#!/usr/bin/python
+#
+# Copyright 2018 Michal Moravec
+# Based on code from Greg Neagle, Timothy Sutton and Per Olofsson
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""See docstring for URLGetter class"""
+
+import os.path
+import subprocess
+
+from autopkglib import Processor, ProcessorError, get_pref, log_err
+
+__all__ = ["URLGetter"]
+
+
+class URLGetter(Processor):
+    """Handles curl HTTP operatios. Server only as superclass. Not for direct use."""
+
+    description = __doc__
+
+    def curl_binary(self):
+        """Returns a path to a curl binary, priority in the order below.
+        Returns None if none found.
+        1. env['CURL_PATH']
+        2. app pref 'CURL_PATH'
+        3. a 'curl' binary that can be found in the PATH environment variable
+        4. '/usr/bin/curl'
+        """
+
+        def is_executable(exe_path):
+            """Is exe_path executable?"""
+            return os.path.exists(exe_path) and os.access(exe_path, os.X_OK)
+
+        if "CURL_PATH" in self.env and is_executable(self.env["CURL_PATH"]):
+            return self.env["CURL_PATH"]
+
+        curl_path_pref = get_pref("CURL_PATH")
+        if curl_path_pref:
+            if is_executable(curl_path_pref):
+                # take a CURL_PATH pref
+                return curl_path_pref
+            else:
+                log_err(
+                    "WARNING: Curl path given in the 'CURL_PATH' preference:'%s' "
+                    "either doesn't exist or is not executable! Falling back "
+                    "to one set in PATH, or /usr/bin/curl." % curl_path_pref
+                )
+
+        for path_env in os.environ["PATH"].split(":"):
+            curlbin = os.path.join(path_env, "curl")
+            if is_executable(curlbin):
+                # take the first 'curl' in PATH that we find
+                return curlbin
+
+        if is_executable("/usr/bin/curl"):
+            # fall back to /usr/bin/curl
+            return "/usr/bin/curl"
+
+        raise ProcessorError("Unable to execute any curl binary")
+
+    def add_curl_common_opts(self, curl_cmd):
+        """Adds request_headers and curl_opts to curl_cmd"""
+        if "request_headers" in self.env:
+            headers = self.env["request_headers"]
+            for header, value in headers.items():
+                curl_cmd.extend(["--header", "%s: %s" % (header, value)])
+        if "curl_opts" in self.env:
+            for item in self.env["curl_opts"]:
+                curl_cmd.extend([item])
+
+    def clear_header(self, header):
+        """Clear header dictionary"""
+        header.clear()
+        header["http_result_code"] = "000"
+        header["http_result_description"] = ""
+
+    def parse_http_protocol(self, line, header):
+        """Parse first HTTP header line"""
+        try:
+            header["http_result_code"] = line.split(None, 2)[1]
+            header["http_result_description"] = line.split(None, 2)[2]
+        except IndexError:
+            pass
+
+    def parse_http_header(self, line, header):
+        """Parse single HTTP header line."""
+        part = line.split(None, 1)
+        fieldname = part[0].rstrip(":").lower()
+        try:
+            header[fieldname] = part[1]
+        except IndexError:
+            header[fieldname] = ""
+
+    def parse_curl_error(self, proc_stderr):
+        """Report Curl failure."""
+        curl_err = ""
+        try:
+            curl_err = proc_stderr.rstrip("\n")
+            curl_err = curl_err.split(None, 2)[2]
+        except IndexError:
+            pass
+
+        return curl_err
+
+    def parse_ftp_header(self, line, header):
+        """Parse single FTP header line."""
+        part = line.split(None, 1)
+        responsecode = part[0]
+        if responsecode == "213":
+            # This is the reply to curl's SIZE command on the file
+            # We can map it to the HTTP content-length header
+            try:
+                header["content-length"] = part[1]
+            except IndexError:
+                pass
+        elif responsecode.startswith("55"):
+            header["http_result_code"] = "404"
+            header["http_result_description"] = line
+        elif responsecode == "150" or responsecode == "125":
+            header["http_result_code"] = "200"
+            header["http_result_description"] = line
+
+    def parse_headers(self, proc_stdout, header):
+        """Parse headers from Curl."""
+        for line in proc_stdout.splitlines():
+            if line.startswith("HTTP/"):
+                self.parse_http_protocol(line, header)
+            elif ": " in line:
+                self.parse_http_header(line, header)
+            elif self.env["url"].startswith("ftp://"):
+                self.parse_ftp_header(line, header)
+            elif line == "":
+                # we got an empty line; end of headers (or curl exited)
+                if header.get("http_result_code") in [
+                    "301",
+                    "302",
+                    "303",
+                    "307",
+                    "308",
+                ]:
+                    # redirect, so more headers are coming.
+                    # Throw away the headers we've received so far
+                    self.clear_header(header)
+
+    def execute_curl(self, curl_cmd):
+        """Executes curl comamnd. Returns stdout, stderr and return code."""
+        proc = subprocess.Popen(
+            curl_cmd,
+            shell=False,
+            bufsize=1,
+            stdin=subprocess.PIPE,
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+        )
+
+        proc_stdout, proc_stderr = proc.communicate()
+        retcode = proc.returncode
+
+        return proc_stdout, proc_stderr, retcode
+
+    def download(self, curl_cmd):
+        """Launches curl returns its output and handles failures."""
+
+        proc_stdout, proc_stderr, retcode = self.execute_curl(curl_cmd)
+
+        if retcode:  # Non-zero exit code from curl => problem with download
+            curl_err = self.parse_curl_error(proc_stderr)
+            raise ProcessorError(
+                "Curl failure: %s (exit code %s)" % (curl_err, retcode)
+            )
+
+        return proc_stdout
+
+    def main(self):
+        pass
+
+
+if __name__ == "__main__":
+    PROCESSOR = URLGetter()
+    PROCESSOR.execute_shell()
--- a/Code/autopkglib/URLGetter.py
+++ b/Code/autopkglib/URLGetter.py
@@ -71,13 +71,11 @@ class URLGetter(Processor):
 
     def add_curl_common_opts(self, curl_cmd):
         """Adds request_headers and curl_opts to curl_cmd"""
-        if "request_headers" in self.env:
-            headers = self.env["request_headers"]
-            for header, value in headers.items():
-                curl_cmd.extend(["--header", "%s: %s" % (header, value)])
-        if "curl_opts" in self.env:
-            for item in self.env["curl_opts"]:
-                curl_cmd.extend([item])
+        for header, value in self.env.get("request_headers", {}).items():
+            curl_cmd.extend(["--header", "%s: %s" % (header, value)])
+
+        for item in self.env.get("curl_opts", []):
+            curl_cmd.extend([item])
 
     def clear_header(self, header):
         """Clear header dictionary"""
--- a/Code/autopkglib/URLGetter.py
+++ b/Code/autopkglib/URLGetter.py
@@ -19,7 +19,7 @@
 import os.path
 import subprocess
 
-from autopkglib import Processor, ProcessorError, get_pref, log_err
+from autopkglib import Processor, ProcessorError, get_pref, is_executable, log_err
 
 __all__ = ["URLGetter"]
 
@@ -37,10 +37,6 @@ class URLGetter(Processor):
         3. a 'curl' binary that can be found in the PATH environment variable
         4. '/usr/bin/curl'
         """
-
-        def is_executable(exe_path):
-            """Is exe_path executable?"""
-            return os.path.exists(exe_path) and os.access(exe_path, os.X_OK)
 
         if "CURL_PATH" in self.env and is_executable(self.env["CURL_PATH"]):
             return self.env["CURL_PATH"]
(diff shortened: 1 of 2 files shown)


Code/autopkglib/URLGetter.py
```diff
@@ -0,0 +1,189 @@
...
+
+__all__ = ["URLGetter"]
+
+
+class URLGetter(Processor):
+    """Handles curl HTTP operatios. Server only as superclass. Not for direct use."""
+
+    description = __doc__
+
+    def curl_binary(self):
```
MichalMMac: `curl_binary()` method was copied from [autopkglib](https://github.com/autopkg/autopkg/blob/15c6307254d4ef1d1dbb8a2648a204a340f5b967/Code/autopkglib/__init__.py#L268.)

Code/autopkglib/URLGetter.py
```diff
@@ -0,0 +1,189 @@
...
+        2. app pref 'CURL_PATH'
+        3. a 'curl' binary that can be found in the PATH environment variable
+        4. '/usr/bin/curl'
+        """
+
+        def is_executable(exe_path):
+            """Is exe_path executable?"""
+            return os.path.exists(exe_path) and os.access(exe_path, os.X_OK)
+
+        if "CURL_PATH" in self.env and is_executable(self.env["CURL_PATH"]):
```
MichalMMac: Added code for getting `curl` path from Processor `CURL_PATH` environment variable:

Code/autopkglib/URLGetter.py
```diff
@@ -0,0 +1,189 @@
...
+                # take the first 'curl' in PATH that we find
+                return curlbin
+
+        if is_executable("/usr/bin/curl"):
+            # fall back to /usr/bin/curl
+            return "/usr/bin/curl"
+
+        raise ProcessorError("Unable to execute any curl binary")
+
+    def add_curl_common_opts(self, curl_cmd):
```
MichalMMac: Adapted from URLDownloader [code] (lines 185..192).(https://github.com/autopkg/autopkg/blob/15c6307254d4ef1d1dbb8a2648a204a340f5b967/Code/autopkglib/URLDownloader.py#L185)

Code/autopkglib/URLGetter.py
```diff
@@ -0,0 +1,189 @@
...
+        for item in self.env.get("curl_opts", []):
+            curl_cmd.extend([item])
+
+    def clear_header(self, header):
+        """Clear header dictionary"""
+        header.clear()
+        header["http_result_code"] = "000"
+        header["http_result_description"] = ""
+
+    def parse_http_protocol(self, line, header):
```
MichalMMac: [Code](https://github.com/autopkg/autopkg/blob/15c6307254d4ef1d1dbb8a2648a204a340f5b967/Code/autopkglib/URLDownloader.py#L229) from `URLDownloader` (lines 229..223).

Code/autopkglib/URLGetter.py
```diff
@@ -0,0 +1,189 @@
...
+        """Parse first HTTP header line"""
+        try:
+            header["http_result_code"] = line.split(None, 2)[1]
+            header["http_result_description"] = line.split(None, 2)[2]
+        except IndexError:
+            pass
+
+    def parse_http_header(self, line, header):
```
MichalMMac: [Code](https://github.com/autopkg/autopkg/blob/15c6307254d4ef1d1dbb8a2648a204a340f5b967/Code/autopkglib/URLDownloader.py#L236) from URLDownloader (lines 236..24).

Code/autopkglib/URLGetter.py
```diff
@@ -0,0 +1,189 @@
...
+        """Parse single HTTP header line."""
+        part = line.split(None, 1)
+        fieldname = part[0].rstrip(":").lower()
+        try:
+            header[fieldname] = part[1]
+        except IndexError:
+            header[fieldname] = ""
+
+    def parse_curl_error(self, proc_stderr):
```
MichalMMac: [Code](https://github.com/autopkg/autopkg/blob/15c6307254d4ef1d1dbb8a2648a204a340f5b967/Code/autopkglib/URLDownloader.py#L285) from URLDownloader (lines 285..292).

Code/autopkglib/URLGetter.py
```diff
@@ -0,0 +1,189 @@
...
+        curl_err = ""
+        try:
+            curl_err = proc_stderr.rstrip("\n")
+            curl_err = curl_err.split(None, 2)[2]
+        except IndexError:
+            pass
+
+        return curl_err
+
+    def parse_ftp_header(self, line, header):
```
MichalMMac: [Code](https://github.com/autopkg/autopkg/blob/15c6307254d4ef1d1dbb8a2648a204a340f5b967/Code/autopkglib/URLDownloader.py#L243) from URLDownloader (lines 243..257).

Code/autopkglib/URLGetter.py
```diff
@@ -0,0 +1,189 @@
...
+            except IndexError:
+                pass
+        elif responsecode.startswith("55"):
+            header["http_result_code"] = "404"
+            header["http_result_description"] = line
+        elif responsecode == "150" or responsecode == "125":
+            header["http_result_code"] = "200"
+            header["http_result_description"] = line
+
+    def parse_headers(self, proc_stdout, header):
```
MichalMMac: `parse_headers()` method is updated version of [code](https://github.com/autopkg/autopkg/blob/15c6307254d4ef1d1dbb8a2648a204a340f5b967/Code/autopkglib/URLDownloader.py#L227) from `URLDownloaded` (lines 227..274).

Functionality in if/else blocks was split into `parse_http_protocol()`, `parse_http_header()`,
`parse_ftp_header))` and `clear_header()` methods.

Code/autopkglib/URLGetter.py
```diff
@@ -0,0 +1,189 @@
...
+                    "302",
+                    "303",
+                    "307",
+                    "308",
+                ]:
+                    # redirect, so more headers are coming.
+                    # Throw away the headers we've received so far
+                    self.clear_header(header)
+
+    def execute_curl(self, curl_cmd):
```
MichalMMac: `execute_curl()` method launches `curl` using `subprocess.Popen` and gets all of its output using `proc.communicate`. This is major change for `URLDownloader`processor and `github` module where reading subprocess output was done using infinite loop. `URLTextSearcher` and `SparkleUpdateInfoProvider` have alredy used `proc.communicate`.

Code/autopkglib/URLGetter.py
nmcspadden: In that case, why separate it out? For obvious reasons, it's completely guaranteed that autopkglib is present on the machine, so why not just use it? I'd rather deduplicate that entirely and just directly import the function from autopkglib.

Code/autopkglib/URLGetter.py
nmcspadden: Although this is fine for the cases here, you might consider just iterating over a list of paths and checking if they're executable there, rather than adding a separate if block for each one.

Code/autopkglib/URLGetter.py
```diff
@@ -0,0 +1,189 @@
...
+                # take a CURL_PATH pref
+                return curl_path_pref
+            else:
+                log_err(
+                    "WARNING: Curl path given in the 'CURL_PATH' preference:'%s' "
+                    "either doesn't exist or is not executable! Falling back "
+                    "to one set in PATH, or /usr/bin/curl." % curl_path_pref
+                )
+
+        for path_env in os.environ["PATH"].split(":"):
```
nmcspadden: I think this is kind of risky. Even though AutoPkg warns against running as root, there's a possibility here that you could run an arbitrary curl without necessarily knowing about it. I think that this should be gated behind an option or preference or something, because I'd want to be very very certain that someone couldn't, say, delete a preference and then trigger what amounts to an arbitrary script (albeit without root privileges).

Code/autopkglib/URLGetter.py
```diff
@@ -0,0 +1,189 @@
...
+    def curl_binary(self):
+        """Returns a path to a curl binary, priority in the order below.
+        Returns None if none found.
+        1. env['CURL_PATH']
+        2. app pref 'CURL_PATH'
+        3. a 'curl' binary that can be found in the PATH environment variable
+        4. '/usr/bin/curl'
+        """
+
+        def is_executable(exe_path):
```
nmcspadden: I'm not a fan of the pattern of a class function defining a function inside itself. If you want to make this a "hidden" function, make it a private function by prefixing it with an underscore, mostly just to ceremonially indicate that it's a private class function, and move it into the class itself.

Code/autopkglib/URLGetter.py
```diff
@@ -0,0 +1,189 @@
...
+            """Is exe_path executable?"""
+            return os.path.exists(exe_path) and os.access(exe_path, os.X_OK)
+
+        if "CURL_PATH" in self.env and is_executable(self.env["CURL_PATH"]):
+            return self.env["CURL_PATH"]
+
+        curl_path_pref = get_pref("CURL_PATH")
+        if curl_path_pref:
+            if is_executable(curl_path_pref):
+                # take a CURL_PATH pref
```
nmcspadden: This comment is unnecessary.

Code/autopkglib/URLGetter.py
nmcspadden: Given that statement, I'm going to review the code in here. To reduce complexity, we should consider changing the code in autopkglib to match the review I'm doing here.

Code/autopkglib/URLGetter.py
```diff
@@ -0,0 +1,189 @@
...
+                )
+
+        for path_env in os.environ["PATH"].split(":"):
+            curlbin = os.path.join(path_env, "curl")
+            if is_executable(curlbin):
+                # take the first 'curl' in PATH that we find
+                return curlbin
+
+        if is_executable("/usr/bin/curl"):
+            # fall back to /usr/bin/curl
```
nmcspadden: I think this comment is also unnecessary, since it's obvious what comes next.

Code/autopkglib/URLGetter.py
```diff
@@ -0,0 +1,189 @@
...
+            return "/usr/bin/curl"
+
+        raise ProcessorError("Unable to execute any curl binary")
```
nmcspadden: To be more precise, the exact error is that we're unable to locate _or_ execute any curl binary.

Code/autopkglib/URLGetter.py
```diff
@@ -0,0 +1,189 @@
...
+        except IndexError:
+            header[fieldname] = ""
+
+    def parse_curl_error(self, proc_stderr):
+        """Report Curl failure."""
+        curl_err = ""
+        try:
+            curl_err = proc_stderr.rstrip("\n")
+            curl_err = curl_err.split(None, 2)[2]
+        except IndexError:
```
nmcspadden: Does this cover all possible error scenarios?

Code/autopkglib/URLGetter.py
```diff
@@ -0,0 +1,189 @@
...
+            curl_cmd,
+            shell=False,
+            bufsize=1,
+            stdin=subprocess.PIPE,
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+        )
+
+        proc_stdout, proc_stderr = proc.communicate()
+        retcode = proc.returncode
```
nmcspadden: It isn't necessary to make this a separate variable; you can return it directly in the return statement.

Code/autopkglib/URLGetter.py
MichalMMac: My intent is to remove `curl_cmd()` function from autopkglib (This was done in my original PR #446 https://github.com/autopkg/autopkg/pull/446/commits/7182befc54bcf1f310f837420053519e45d712bb).

Arguments:

1. All `curl` related code should be in `URLGetter`.
2. No one in autopkg org is using  `curl_cmd()` in their code therefore we should be able to remove it wihout any inpact.

Code/autopkglib/URLGetter.py
MichalMMac: This 4bece62 ok? 

Or should I do it in single line ~ `return proc.communicate() + (proc.returncode, )`?

Code/autopkglib/URLGetter.py
MichalMMac: This code was moved and slightly adapted from [URLDownloader](https://github.com/autopkg/autopkg/blob/15c6307254d4ef1d1dbb8a2648a204a340f5b967/Code/autopkglib/URLDownloader.py#L291).

Do you have an idea how should it be improved?

Code/autopkglib/URLGetter.py
MichalMMac: This comment is in the current master https://github.com/autopkg/autopkg/blob/15c6307254d4ef1d1dbb8a2648a204a340f5b967/Code/autopkglib/__init__.py#L268

I'll happily remove them ;-) See ab7f865

Code/autopkglib/URLGetter.py
MichalMMac: See 1396150

Code/autopkglib/URLGetter.py
MichalMMac: Commented in https://github.com/autopkg/autopkg/pull/544#discussion_r307995642

Code/autopkglib/URLGetter.py
MichalMMac: This code was taken as is from https://github.com/autopkg/autopkg/blob/15c6307254d4ef1d1dbb8a2648a204a340f5b967/Code/autopkglib/__init__.py#L268

My guess is original intent was to reuse the code within `curl_cmd` but nowhere else.

IMO proper thing to do is to move `is_executable()` to `autopkglib` as a standalone function. See 118683f

Code/autopkglib/URLGetter.py
MichalMMac: This code is already in the master. https://github.com/autopkg/autopkg/blob/15c6307254d4ef1d1dbb8a2648a204a340f5b967/Code/autopkglib/__init__.py#L291

Do you think we should deal with this in this PR? If so what do you suggest? Preference like `find_curl_in_path` -> if true run this code?

MichalMMac:
> I like the work here done overall, but I hate duplicating code. I think we should factor out the common code and keep it in autopkglib, and then just import it directly from there.

Please clarify. Do you think we should abandon `URLGetter` idea altogether and move all this code to `autopkglib` OR was this comment only about `curl_cmd ` -> `curl_binary` function/method duplication (I explain my intent in https://github.com/autopkg/autopkg/pull/544#discussion_r307969156)?

Code/autopkglib/URLGetter.py
MichalMMac: Interesting idea. Would require to rewrite `curl_binary()` method. I am going to think about this. My only concern is result being more compact but less human readable.

Code/autopkglib/URLGetter.py
nmcspadden: I'm sort of of the opposite opinion, that curl_cmd() should stay in autopkglib and be referenced here, but honestly don't see a real need to actually do that. So I'm satisfied with this methodology.

Code/autopkglib/URLGetter.py
nmcspadden: I'm okay with copying it as existing from autopkglib, but it's something I'm making a note to investigate further later.

Code/autopkglib/URLGetter.py
nmcspadden: I'm happy with that change.

Code/autopkglib/URLGetter.py
nmcspadden: What you've done is perfectly fine.

Code/autopkglib/URLGetter.py
nmcspadden: Yeah, this is fine for now.

Code/autopkglib/URLGetter.py
```diff
@@ -0,0 +1,181 @@
...
+import os.path
+import subprocess
+
+from autopkglib import Processor, ProcessorError, get_pref, is_executable, log_err
+
+__all__ = ["URLGetter"]
+
+
+class URLGetter(Processor):
+    """Handles curl HTTP operatios. Server only as superclass. Not for direct use."""
```
nmcspadden: """Handles curl HTTP operations. Serves only as superclass

Code/autopkglib/URLGetter.py
```diff
@@ -0,0 +1,181 @@
...
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+        )
+
+        proc_stdout, proc_stderr = proc.communicate()
+
+        return proc_stdout, proc_stderr, proc.returncode
+
+    def download(self, curl_cmd):
+        """Launches curl returns its output and handles failures."""
```
nmcspadden:
```
"""Launches curl, returns its output, and handles failures."""
```

Code/autopkglib/URLGetter.py
MichalMMac: Thanks for catching this. See 842485d

Code/autopkglib/URLGetter.py
MichalMMac: Thanks for catching this. See 842485d

nmcspadden:
@MichalMMac please migrate all of these to the new branch:
https://github.com/autopkg/autopkg/tree/new_urlgetter

MichalMMac:
@nmcspadden I changed the target of all my URLGetter related PRs to `new_urlgetter` branch.

MichalMMac:
@nmcspadden How should we proceed?

nmcspadden:
Now that this has been merged into a secondary branch, it needs to go through a bunch of testing to make sure it's behaving as expected. Once I'm satisfied it doesn't break any workflows or interfere with any normal operations, I'll merge it into master.

MichalMMac:
@nmcspadden Agreed. Where should we discuss problems/changes? New GitHub issues?

nmcspadden:
Yes, new issues should be filed for each problem that needs to be addressed. It's easier to refer to individual issues rather than larger "master issues" that combine several things at once.

PR #547, opened 2019-07-04
SparkleUpdateInfoProvider refactoring using URLGetter

MichalMMac:
This PR was split from #446. It should be merged **after** PR #544 (`URLGetter`). Branch `urlgetter-sparkleupdateinfoprovider` includes code from branch `urlgetter` in order for `SparkleUpdateInfoProvider` to work. `URLGetter` should be reviewed in #544.

`URLTextSearcher` was refactored in following way:

- `SparkleUpdateInfoProvider` is now subclass of `URLGetter`
- Functionality like executing `curl` was moved to `URLGetter`
- Some of the code from `get_feed_data()` was split into methods `build_url()`, `determine_version()`, `parse_feed_data`.
- Some of the code from `main()` was split into methods `handle_pkginfo()`
- Some local variables are now instance variables
- `SparkleUpdateInfoProvider` now supports `curl_opts` variable

### DIFF
--- a/Code/autopkglib/SparkleUpdateInfoProvider.py
+++ b/Code/autopkglib/SparkleUpdateInfoProvider.py
@@ -321,10 +321,7 @@ class SparkleUpdateInfoProvider(URLGetter):
             return
 
         # handle custom xmlns and version attributes
-        if "alternate_xmlns_url" in self.env:
-            self.xmlns = self.env["alternate_xmlns_url"]
-        else:
-            self.xmlns = DEFAULT_XMLNS
+        self.xmlns = self.env.get("alternate_xmlns_url", DEFAULT_XMLNS)
 
         data = self.get_feed_data(self.env.get("appcast_url"))
         items = self.parse_feed_data(data)
@@ -340,10 +337,7 @@ class SparkleUpdateInfoProvider(URLGetter):
         pkginfo = self.handle_pkginfo(latest)
 
         self.env["url"] = latest["url"]
-        if latest.get("human_version"):
-            self.env["version"] = latest["human_version"]
-        else:
-            self.env["version"] = latest["version"]
+        self.env["version"] = latest.get("human_version", latest["version"])
         self.output("Found URL %s" % self.env["url"])
         self.env["additional_pkginfo"] = pkginfo
 
--- a/Code/autopkglib/SparkleUpdateInfoProvider.py
+++ b/Code/autopkglib/SparkleUpdateInfoProvider.py
@@ -142,15 +142,10 @@ class SparkleUpdateInfoProvider(URLGetter):
 
     def fetch_content(self, url, headers=None):
         """Returns content retrieved by curl, given a url and an optional
-        dictionary of header-name/value mappings. Logic here borrowed from
-        URLTextSearcher processor."""
+        dictionary of header-name/value mappings."""
 
         curl_cmd = self.prepare_curl_cmd(url, headers)
-        try:
-            content = super(SparkleUpdateInfoProvider, self).download(curl_cmd)
-        except ProcessorError:
-            raise ProcessorError("Could not retrieve URL: %s" % url)
-
+        content = super(SparkleUpdateInfoProvider, self).download(curl_cmd)
         return content
 
     def get_feed_data(self, url):
--- a/Code/autopkglib/SparkleUpdateInfoProvider.py
+++ b/Code/autopkglib/SparkleUpdateInfoProvider.py
@@ -1,5 +1,6 @@
 #!/usr/bin/python
 #
+# Refactoring 2018 Michal Moravec
 # Copyright 2013-2016 Timothy Sutton
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
@@ -17,14 +18,14 @@
 """See docstring for SparkleUpdateInfoProvider class"""
 
 import os
-import subprocess
 import urllib
 import urlparse
 from distutils.version import LooseVersion
 from operator import itemgetter
 from xml.etree import ElementTree
 
-from autopkglib import Processor, ProcessorError
+from autopkglib import ProcessorError
+from autopkglib.URLGetter import URLGetter
 
 __all__ = ["SparkleUpdateInfoProvider"]
 
@@ -32,7 +33,7 @@ DEFAULT_XMLNS = "http://www.andymatuschak.org/xml-namespaces/sparkle"
 SUPPORTED_ADDITIONAL_PKGINFO_KEYS = ["description", "minimum_os_version"]
 
 
-class SparkleUpdateInfoProvider(Processor):
+class SparkleUpdateInfoProvider(URLGetter):
     """Provides URL to the highest version number or latest update."""
 
     description = __doc__
@@ -58,6 +59,12 @@ class SparkleUpdateInfoProvider(Processor):
                 "Alternate URL for the XML namespace, if the "
                 "appcast is using an alternate one. Defaults to "
                 "that used for 'vanilla' Sparkle appcasts."
+            ),
+        },
+        "curl_opts": {
+            "required": False,
+            "description": (
+                "Optional array of options to include with the download " "request."
             ),
         },
         "pkginfo_keys_to_copy_from_sparkle_feed": {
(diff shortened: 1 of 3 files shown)


Code/autopkglib/SparkleUpdateInfoProvider.py
```diff
@@ -118,27 +125,88 @@ class SparkleUpdateInfoProvider(Processor):
...
+        if "appcast_query_pairs" in self.env:
+            queries = self.env["appcast_query_pairs"]
+            new_query = urllib.urlencode([(k, v) for (k, v) in queries.items()])
+            scheme, netloc, path, _, frag = urlparse.urlsplit(url)
+            url = urlparse.urlunsplit((scheme, netloc, path, new_query, frag))
+
+        data = self.fetch_content(url, headers=self.env.get("appcast_request_headers"))
+        return data
+
+    def build_url(self, enclosure):
```
MichalMMac: Split from `get_feed_data()` [lines](https://github.com/autopkg/autopkg/blob/15c6307254d4ef1d1dbb8a2648a204a340f5b967/Code/autopkglib/SparkleUpdateInfoProvider.py#L186) 186..194

Code/autopkglib/SparkleUpdateInfoProvider.py
```diff
@@ -118,27 +125,88 @@ class SparkleUpdateInfoProvider(Processor):
...
+        if self.env.get("urlencode_path_component", True):
+            encoded_path = urllib.quote(url_bits.path)
+        else:
+            encoded_path = url_bits.path
+        built_url = url_bits.scheme + "://" + url_bits.netloc + encoded_path
+        if url_bits.query:
+            built_url += "?" + url_bits.query
+        return built_url
+
+    def determine_version(self, enclosure, url):
```
MichalMMac: Split from `get_feed_data()` [lines](https://github.com/autopkg/autopkg/blob/15c6307254d4ef1d1dbb8a2648a204a340f5b967/Code/autopkglib/SparkleUpdateInfoProvider.py#L196) 196..217

Code/autopkglib/SparkleUpdateInfoProvider.py
```diff
@@ -118,27 +125,88 @@ class SparkleUpdateInfoProvider(Processor):
...
+            for delimiter in ["_", "-"]:
+                if delimiter in filename:
+                    version = filename.split(delimiter)[-1]
+                    break
+        # if we still found nothing, fail
+        if version is None:
+            raise ProcessorError("Can't extract version info from item in feed!")
+        return version
+
+    def parse_feed_data(self, data):
```
MichalMMac: Split from `get_feed_data()` some of the [lines](https://github.com/autopkg/autopkg/blob/15c6307254d4ef1d1dbb8a2648a204a340f5b967/Code/autopkglib/SparkleUpdateInfoProvider.py#L170) in range 170..233

Code/autopkglib/SparkleUpdateInfoProvider.py
```diff
@@ -153,20 +221,6 @@ def get_feed_data(self, url):
...
         by whoever calls this function."""
 
-        # handle custom xmlns and version attributes
-        if "alternate_xmlns_url" in self.env:
-            xmlns = self.env["alternate_xmlns_url"]
-        else:
-            xmlns = DEFAULT_XMLNS
-
-        # query string
-        if "appcast_query_pairs" in self.env:
```
MichalMMac: Now in `get_feed_data()`

Code/autopkglib/SparkleUpdateInfoProvider.py
```diff
@@ -153,20 +221,6 @@ def get_feed_data(self, url):
         metadata we're never going to use. If it's a URL, this must be handled
         by whoever calls this function."""
 
-        # handle custom xmlns and version attributes
```
MichalMMac: Now in `main()`

Code/autopkglib/SparkleUpdateInfoProvider.py
```diff
@@ -181,59 +235,64 @@ def get_feed_data(self, url):
             enclosure = item_elem.find("enclosure")
             if enclosure is not None:
                 item = {}
-                # URL-quote the path component to handle spaces, etc.
```
MichalMMac: Now in `build_url()`

Code/autopkglib/SparkleUpdateInfoProvider.py
```diff
@@ -181,59 +235,64 @@ def get_feed_data(self, url):
...
-                if self.env.get("urlencode_path_component", True):
-                    encoded_path = urllib.quote(url_bits.path)
-                else:
-                    encoded_path = url_bits.path
-                built_url = url_bits.scheme + "://" + url_bits.netloc + encoded_path
-                if url_bits.query:
-                    built_url += "?" + url_bits.query
-                item["url"] = built_url
-
-                item["version"] = enclosure.get("{%s}version" % xmlns)
```
MichalMMac: now in `determine_version()`

Code/autopkglib/SparkleUpdateInfoProvider.py
```diff
@@ -271,36 +337,7 @@ def compare_version(this, that):
                 % latest["human_version"]
             )
 
-        pkginfo = {}
```
MichalMMac: Now in `handle_pkginfo()`.

Code/autopkglib/SparkleUpdateInfoProvider.py
```diff
@@ -181,59 +235,64 @@ def get_feed_data(self, url):
...
                 if description_elem is not None:
                     item["description_url"] = description_elem.text.strip()
+
                 if item_elem.find("description") is not None:
                     item["description_data"] = item_elem.find("description").text
                 versions.append(item)
 
         return versions
 
+    def handle_pkginfo(self, latest):
```
MichalMMac: Code from `main()`

Code/autopkglib/SparkleUpdateInfoProvider.py
```diff
@@ -118,27 +125,88 @@ class SparkleUpdateInfoProvider(Processor):
...
+    def prepare_curl_cmd(self, url, headers=None):
+        """Assemble curl command and return it."""
+
+        curl_cmd = [
+            super(SparkleUpdateInfoProvider, self).curl_binary(),
+            "--location",
+            "--compressed",
+        ]
+        super(SparkleUpdateInfoProvider, self).add_curl_common_opts(curl_cmd)
+        if headers:
```
nmcspadden: This logic is already handled in URLGetter.py's `add_curl_common_opts` - why do it here again?

Code/autopkglib/SparkleUpdateInfoProvider.py
```diff
@@ -118,27 +125,88 @@ class SparkleUpdateInfoProvider(Processor):
...
-                    cmd.extend(["--header", "%s: %s" % (header, value)])
-            cmd.append(url)
-            proc = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
-            (data, stderr) = proc.communicate()
-            if proc.returncode:
-                raise ProcessorError("Could not retrieve URL %s: %s" % (url, stderr))
-        except OSError:
+            content = super(SparkleUpdateInfoProvider, self).download(curl_cmd)
+        except ProcessorError:
             raise ProcessorError("Could not retrieve URL: %s" % url)
```
nmcspadden: Why bother catching and then raising the same exception? The download() function already raises an exception that reports a curl failure. I don't really see the benefit of catching it just to reraise with different phrasing.

Code/autopkglib/SparkleUpdateInfoProvider.py
```diff
@@ -181,59 +235,64 @@ def get_feed_data(self, url):
...
-                # if we still found nothing, fail
-                if item["version"] is None:
-                    raise ProcessorError(
-                        "Can't extract version info from item in feed!"
-                    )
+                item["url"] = self.build_url(enclosure)
+                item["version"] = self.determine_version(enclosure, item["url"])
 
-                human_version = enclosure.get("{%s}shortVersionString" % xmlns)
+                human_version = enclosure.get("{%s}shortVersionString" % self.xmlns)
```
nmcspadden: Not sure I'd necessarily state that this is the "human version" because it depends heavily on the vendor, but not a blocker.

Code/autopkglib/SparkleUpdateInfoProvider.py
```diff
@@ -181,59 +235,64 @@ def get_feed_data(self, url):
...
+                elif "description_data" in latest.keys():
+                    description = (
+                        "<html><body>" + latest["description_data"] + "</html></body>"
+                    )
+                else:
+                    description = ""
+                pkginfo["description"] = description.decode("UTF-8")
+
+            if "minimum_os_version" in sparkle_pkginfo_keys:
+                if latest.get("minimum_os_version") is not None:
```
nmcspadden: None is already not truth, so you can compress this:
```
if latest.get("minimum_os_version")
```

Yes, I know and see that this construction is everywhere, but it's something we'll have to change on our next revision.

Code/autopkglib/SparkleUpdateInfoProvider.py
MichalMMac: Similiar reply as in https://github.com/autopkg/autopkg/pull/548#discussion_r319856830

These headers are not from Processor `self.env.get("request_headers")`. `SparkleUpdateInfoProvider` does not have `request_headers` input variable. Instead there is a `appcast_request_headers` variable which is specific to `SparkleUpdateInfoProvider` only.

We could change the name of the variable but that would break many recipes https://github.com/search?q=appcast_request_headers&type=Code.

Code/autopkglib/SparkleUpdateInfoProvider.py
MichalMMac: What alternate name do you suggest?

Code/autopkglib/SparkleUpdateInfoProvider.py
MichalMMac: Ok. I remove the code in 94022d3.

Code/autopkglib/SparkleUpdateInfoProvider.py
MichalMMac: What if `minimum_os_version` is zero?

I think `is not None` construction is there for this purpose thus I did not change it in the refactoring

PR #567, opened 2019-10-01
Use shallow clones/pulls for recipe repos

williamtheaker:
I'm starting a thread on autopkg-discuss about this change, but the general idea is to use [shallow clones][1] and pulls for recipe repos to speed up `repo-add` and `repo-update`, as well as reduce disk space usage.

[1]: https://git-scm.com/docs/git-clone#Documentation/git-clone.txt---depthltdepthgt

### DIFF
--- a/Code/autopkg
+++ b/Code/autopkg
@@ -604,7 +604,7 @@ def get_recipe_repo(git_path):
     else:
         log("Attempting git clone...")
         try:
-            log(run_git(["clone", git_path, dest_dir]))
+            log(run_git(["clone", "--depth", "1", git_path, dest_dir]))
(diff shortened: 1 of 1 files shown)


nmcspadden:
Can you verify that this doesn't affect trust info verification? 

1. Make an override for one of the recipes in a git repo (that you control)
2. Update the recipe in the git repo so that trust is broken
3. Try running again 

Does it affect the output at all?

nmcspadden:
This hasn't been updated since October, and none of my questions were answered. If you wish to pursue this, please reopen, rebase, re-test, and ensure that the output remains the same.

PR #580, opened 2019-11-14
WIP: Py2 to 3 fixes

MichalMMac:
Fixes to get `URLGetter` code working after py2-to-3 conversion. Currently WIP.

- Popen in `URLGetter` is set to text mode by default 5e3c35c. This is handy when expecting **text* output from subprocess we want to parse/searchh. I made text mode configurable 7c05276 because sometimes we might need bytes output. Popen in `CodeSignatureVerifier` set to text mode only 845337e
- URLDownloader str/bytes fixes beac4a5, 1a1777f
- Fixed SparkleUpdateInfoProvider urllib code 3614ee4
- Fixed SparkleUpdateInfoProvider missing curly brackets bug introduced in py2-to-3 branch
- SparkleUpdateInfoProvider  Sort function converted to Python3 e314375 **Plese take a look** at lambda function solution. It works for me but I don't use lambda very often.
- DmgMounter str/bytes fixes f55d3a0
- URLDownloader calls `curl` with `--fail` only when downlading something. No need for `--fail` when prefetching filname b01d69a

**Update: 17.11.**:

- dbb5acd sets text or binary mode for all Popen calls. **Please take a look**
- beac4a5 and 1a1777f fix autopkgserver socket encoding issues

### DIFF
--- a/Code/autopkg
+++ b/Code/autopkg
@@ -548,7 +548,11 @@ def run_git(git_options_and_arguments, git_directory=None):
     cmd.extend(git_options_and_arguments)
     try:
         proc = subprocess.Popen(
-            cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE, cwd=git_directory
+            cmd,
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+            cwd=git_directory,
+            text=True,
         )
         (cmd_out, cmd_err) = proc.communicate()
     except OSError as err:
@@ -559,7 +563,7 @@ def run_git(git_options_and_arguments, git_directory=None):
     if proc.returncode != 0:
         raise GitError(f"ERROR: {cmd_err}")
     else:
-        return cmd_out.decode("UTF-8")
+        return cmd_out
 
 
 def get_recipe_repo(git_path):
--- a/Code/autopkglib/CodeSignatureVerifier.py
+++ b/Code/autopkglib/CodeSignatureVerifier.py
@@ -185,7 +185,7 @@ class CodeSignatureVerifier(DmgMounter):
         """
         process = ["/usr/sbin/pkgutil", "--check-signature", path]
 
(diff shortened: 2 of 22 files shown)


MichalMMac:
**TODO** Known issue. After sucessfull package install Installer processor tries to close the socket.

```
Installer: Disconnecting
  File "/Library/AutoPkg/autopkglib/__init__.py", line 673, in process
  File "/Library/AutoPkg/autopkglib/__init__.py", line 480, in process
    self.main()
  File "/Library/AutoPkg/autopkglib/Installer.py", line 181, in main
    self.install()
  File "/Library/AutoPkg/autopkglib/Installer.py", line 127, in install
    self.disconnect()
  File "/Library/AutoPkg/autopkglib/Installer.py", line 177, in disconnect
    self.socket.close()
  File "/Users/Shared/Python3/Python.framework/Versions/3.7/lib/python3.7/socket.py", line 420, in close
    self._real_close()
  File "/Users/Shared/Python3/Python.framework/Versions/3.7/lib/python3.7/socket.py", line 414, in _real_close
    _ss.close(self)
OSError: [Errno 9] Bad file descriptor
  File "/Library/AutoPkg/autopkglib/__init__.py", line 673, in process
[Errno 9] Bad file descriptor
```

nmcspadden:
I'm merging this as-is because it's way easier to continue development this way. Thank you so much for this work, it's a huge boost to have someone else also contributing.

Looking at your changes to the subprocess.Popen() calls, though, makes me think that we should just switch every single one of them to subprocess.run() instead, introduced in 3.5, with the _text_ feature introduced in 3.7. Since I'm declaring AutoPkg to require python 3.7.5 at a minimum, this seems like an effective way to reduce many unnecessary lines of code.

MichalMMac:
Next time we could at least remove the `WIP:` from the PR title 😄 

`subprocess.run()` look like intersting alternative. I'll take a look in a few days.

PR #583, opened 2019-11-21
URLGetter py2 upgrades

MichalMMac:
Changes to the `URLGetter` and its subclasses to enable `autopkg/recipes` py23 compatibiilty upgrade with use of `URLGetter`.

Similiar changes future ported to py2-to-3 branch in PR #584

Changes:

- Oriignal `download()` method was renamed to `download_with_curl()` so we can use **download** for simpler uses. b1e449b
- `URLGetter` now has `prepare_curl_cmd()` with *default* `curl` command e45ce95. This is used by new `download()` method and subclasses. 71296de
- We now use inherited methods instead of calling `super()` everywhere. 3a17705
- New simple `download()` method 32267dc for use cases where we don't want to specify `curl` command. It takes http headers too 7b47aa1. We don't use `text` argument in python2 code but it's there for recipe compatibility with python3 version.

### DIFF
--- a/Code/autopkglib/URLGetter.py
+++ b/Code/autopkglib/URLGetter.py
@@ -190,7 +190,7 @@ class URLGetter(Processor):
 
         return proc_stdout
 
-    def download(self, url, headers=None):
+    def download(self, url, headers=None, text=False):
         """Download content with default curl options"""
 
         curl_cmd = self.prepare_curl_cmd()
--- a/Code/autopkglib/URLGetter.py
+++ b/Code/autopkglib/URLGetter.py
@@ -47,9 +47,9 @@ class URLGetter(Processor):
                 return curl_path_pref
             else:
                 log_err(
-                    "WARNING: curl path given in the 'CURL_PATH' preference:'%s' "
+                    "WARNING: curl path given in the 'CURL_PATH' preference:'{}' "
                     "either doesn't exist or is not executable! Falling back "
-                    "to one set in PATH, or /usr/bin/curl." % curl_path_pref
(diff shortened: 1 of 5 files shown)


nmcspadden:
new_urlgetter was merged into master, so please rebase your branch on master (because it's currently out of date, from what I can tell). I thought I deleted the new_urlgetter branch entirely on the github remote, but I apparently did not do so successfully.

MichalMMac:
`new_urlgetter` branch pointed to the same as the `master`. Target now changed to `master`.

Code/autopkglib/URLGetter.py
```diff
@@ -62,10 +62,19 @@ def curl_binary(self):
...
 
+    def prepare_curl_cmd(self):
+        """Assemble basic curl command and return it."""
+        return [self.curl_binary(), "--compressed", "--location"]
+
+    def add_curl_headers(self, curl_cmd, headers):
+        """Add headers to curl_cmd"""
+        if headers:
+            for header, value in headers.items():
+                curl_cmd.extend(["--header", "%s: %s" % (header, value)])
```
nmcspadden: Use `.format()` instead of string interpolation here, because it'll be easier to transition in the future.

MichalMMac: Ok. Will do.

MichalMMac: Pushed 5 new commits replacing string interpolation with format in `URLGetter` and subclasses.

nmcspadden:
I tested this and it works for all recipes currently in the Py2-to-3 autopkg-recipes branch (as of commit 25518b9b761f943d4c567fcb0944561b32ca3b89)

PR #586, opened 2019-11-21
Update DmgMounter.py

gregneagle:
In DMGMounter's dmg_has_sla do not return False if hdiutil has stderr output. Addresses an issue with mounting APFS dmgs with SLA/EULAs.


gregneagle:
See https://github.com/munki/munki/blob/master/code/client/munkilib/dmgutils.py#L59-L80 for comparison.

PR #591, opened 2019-12-03
Create unit test for FileCreator

homebysix:


nmcspadden:
Are you still intending to add to this, or is it ready for review now? Your skeleton is fine, but the unit test itself doesn't do much.

homebysix:
Let me take a shot at the mock_open, and if I can get that sorted I’ll remove the WIP.

homebysix:
OK, I think I got it. Whether it's a useful test is debatable.

```
% python3 -m unittest tests.test_filecreator -v
test_file_content (tests.test_filecreator.TestFileCreator)
The file created by the processor should have the expected contents. ... ok

----------------------------------------------------------------------
Ran 1 test in 0.008s

OK
```

nmcspadden:
This is good! 

In a future PR, I think we should add a few more tests:
* Opening a bad file path should trigger a ProcessorError
* Opening a good file path should _not_ trigger any error
* Test the file mode is created as expected
* Failing to set the file mode, if defined, should trigger a ProcessorError

nmcspadden:
Additionally, we should ensure we have the correct copyright headers on all files in the repo.

PR #596, opened 2019-12-12
URLTextSearcher unit tests

MichalMMac:
- Small refactoring to better structure the code
- Unit tests for URLTextSearcher covering all? its features

### DIFF
--- a/Code/autopkglib/URLTextSearcher.py
+++ b/Code/autopkglib/URLTextSearcher.py
@@ -81,6 +81,9 @@ class URLTextSearcher(URLGetter):
 
     description = __doc__
 
+    match_message = "Found matching text"
+    no_match_message = "No match found on URL"
+
     def prepare_curl_cmd(self):
         """Assemble curl command and return it."""
         curl_cmd = super().prepare_curl_cmd()
@@ -103,7 +106,7 @@ class URLTextSearcher(URLGetter):
         match = re_pattern.search(content)
 
         if not match:
-            raise ProcessorError(f"No match found on URL: {self.env['url']}")
+            raise ProcessorError(f"{self.no_match_message}: {self.env['url']}")
 
         # return the last matched group with the dict of named groups
         return (match.group(match.lastindex or 0), match.groupdict())
@@ -125,7 +128,7 @@ class URLTextSearcher(URLGetter):
         self.output_variables = {}
         for key in groupdict.keys():
             self.env[key] = groupdict[key]
-            self.output(f"Found matching text ({key}): {self.env[key]}")
+            self.output(f"{self.match_message} ({key}): {self.env[key]}")
             self.output_variables[key] = {
                 "description": "Matched regular expression group"
             }
--- a/Code/autopkglib/URLTextSearcher.py
+++ b/Code/autopkglib/URLTextSearcher.py
@@ -29,7 +29,7 @@ class URLTextSearcher(URLGetter):
     """Downloads a URL using curl and performs a regular expression match
     on the text.
 
-    Requires version 0.2.9."""
+    Requires version 1.4."""
 
     input_variables = {
         "re_pattern": {
--- a/Code/autopkglib/URLTextSearcher.py
+++ b/Code/autopkglib/URLTextSearcher.py
@@ -88,13 +88,18 @@ class URLTextSearcher(URLGetter):
         curl_cmd.append(self.env["url"])
         return curl_cmd
 
-    def re_search(self, content):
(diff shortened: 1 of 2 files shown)


Code/autopkglib/URLTextSearcher.py
```diff
@@ -81,24 +81,32 @@ class URLTextSearcher(URLGetter):
...
+
+    def re_search(self, content):
+        """Search for re_pattern in content"""
+
+        re_pattern = re.compile(self.env["re_pattern"], flags=self.prepare_re_flags())
         match = re_pattern.search(content)
 
         if not match:
-            raise ProcessorError(f"No match found on URL: {self.env['url']}")
+            raise ProcessorError(f"{self.no_match_message}: {self.env['url']}")
```
nmcspadden: Not sure I really get the point of this, but okay.

nmcspadden:
Can you please provide the output of running the unit tests in the PR as well?

nmcspadden:
Nice work overall! Just please include the output and I'll merge it.

Code/autopkglib/URLTextSearcher.py
MichalMMac: `self.no_match_message` variable is used in the test class. When there is no match `ProcessorError` gets raised. Since it's quite generic exception I check for `No match found on URL`. If someone decide to reword this string test breaks. So I was thinking to put into variable I can reuse.

I am not sure if this is a good idea or anti pattern. If you think it's bad approach I can drop the 1331358 commit and update the test.

MichalMMac:
This output? :

```
extSearcher
test_case_insensitive_re_pattern_flag (tests.test_urltextsearcher.TestURLTextSearcher)
With re.IGNORECASE flag difference in character case should match. ... ok
--
test_case_sensitive_re_pattern_default (tests.test_urltextsearcher.TestURLTextSearcher)
Difference in character case should not produce a match by default. ... ok
--
test_found_a_match (tests.test_urltextsearcher.TestURLTextSearcher)
If processor finds a match, it should be in the env. ... ok
--
test_match_last_unnamed_group (tests.test_urltextsearcher.TestURLTextSearcher)
Only last unnamed re group should matchwhen multiple groups used. ... ok
--
test_match_multiple_named_groups (tests.test_urltextsearcher.TestURLTextSearcher)
Processor should return output variable for each named re group. ... ok
--
test_match_named_group_with_output_var (tests.test_urltextsearcher.TestURLTextSearcher)
Only named group with name equal to result_output_var_name should match. ... ok
--
test_no_fail_if_good_env (tests.test_urltextsearcher.TestURLTextSearcher)
--
test_not_found_a_match (tests.test_urltextsearcher.TestURLTextSearcher)
If processor does not find a match, ProcessorError should be raised. ... ok
```

BTW how can I run single test file? I am running unittests like this: `python3 -m unittest discover -v Code`. When I try to run single test file only there is a problem with autopkglib import.

Code/autopkglib/URLTextSearcher.py
nmcspadden: > If someone decide to reword this string test breaks. So I was thinking to put into variable I can reuse.

That's a fair argument, but if that's the case, we should codify it even more:
* URLTextSearcher should put that output text into a constant
* URLTextSearcher should raise an exception with the contents of the constant when no match is found
* This unit test should import that constant, and use that for the test

nmcspadden:
I've merged this, but I still think we should codify the message string into a constant; that's worth a separate PR.

MichalMMac:
What exactly do you mean by codifing the message into a constant?
Should I use the constant module? Or should I make the variable name upper case?

nmcspadden:
Sorry, I should have been more precise. 

Create an uppercase-variable at the top of the module, such as:
```
NO_MATCH_MESSAGE = "No match found on URL"
MATCH_MESSAGE = ""Found matching text"
```

Then the unit test can simply do this:
```
from autopkglib.URLTextSearcher import URLTextSearcher, NO_MATCH_MESSAGE, MATCH_MESSAGE
```

And use it accordingly.

MichalMMac:
Is using the variable outside the class better approach than class variable? Class variables are often used as "constants".

I can change it to upper case though.

nmcspadden:
Is it "better"? Probably debatable, maybe even more of a style question, but I like keeping it in a module constant because you can use it in more places without also having to incur the overhead of importing the main processor class itself.

However, literally nothing about this is really tuned for high performance, so the distinction is largely a preference I'm biased towards.

Issue #432, opened 2018-09-24
URLDownloader Feature Request: Assign filename from redirected URL, not original one


jazzace:
I found this specifically when working with the Microsoft Office recipes that download full installers (@arubdesu, @rtrouton). If you use Microsoft's FWLink URLs (e.g., https://go.microsoft.com/fwlink/?linkid=871748), they redirect to the current URL by returning a HTTP/1.1 302 status. AutoPkg correctly parses that, gets the new URL, and downloads the correct item. Based on the code on Line 139 of URLDownloader, the download is assigned the name of whatever is after the final slash in the original URL (in my example, `?linkid=871748`).

If you enter the same URL in a web browser, the resulting download is named in a more expected manner, that is, it uses the filename being downloaded (which is equivalent to what is after the final slash in the resolved/final URL, which in the above example is `Microsoft_Word_16.16.18091001_Installer.pkg`).

In the case where URLDownloader picks up a new URL due to a 301, 302, or 303 redirect (Lines 250-256), it would better match user expectations (and could potentially be more useful) if the output filename was changed to what the final URL provides. Perhaps this is something that could be considered if/when the URLGetter class suggested in Issue #416 is developed.

MichalMMac:
Look at my refactor branch here https://github.com/MichalMMac/autopkg/tree/curlrefactor

This issue is covered in this commit https://github.com/MichalMMac/autopkg/commit/b90d04f627d5f04de3b1f598de2dc4ff17f61d21.

Feel free to test my refactoring attempt and discuss in autopkg google [group](https://groups.google.com/forum/#!topic/autopkg-discuss/i4FswIat6GA).

n8felton:
I believe this is also covered in #254.

MichalMMac:
I added `rename_after_redirect` switch set to `False` by default making this functionality optional. https://github.com/MichalMMac/autopkg/commit/88d5232f001bf07d12857faf43da9962ba60e20e

MichalMMac:
It's not as simple as I though. My previous solution broke etag checking. Since renamed filename is not known before download it was downloaded every time.

Updated solution https://github.com/MichalMMac/autopkg/commit/c0024278099cf3841a776c9005a7426ca0df9802 :

1. Downloads the file last part 
2. Names it with last part of `url` variable or `filename` variable (when set)
3. Instead of just renaming it to filename discovered from Location HTTP header it creates a hard link.

[...]

PR #599, opened 2019-12-14
Add debug output to URLDownloader.prefetch_filename()

MichalMMac:
Adds `verbose_level=2` output to `prefetch_filename()` method printing out what filename was found in which header. This should help autopkg users to debug their recipes.

Requested here https://github.com/autopkg/autopkg/issues/432#issuecomment-559154131


nmcspadden:
In general, it's helpful to provide sample output in the PR itself when you are changing the text that admins observe, to give viewers (and reviewers) an expectation in context.

PR #600, opened 2019-12-17
Versioner changes and unittests

MichalMMac:
- Code for loading plists was moved into base `Processor. self.load_plist_from_file()` method (da3a354 and 428831a) This method could be used in other processors. Ouput:
    - If path to plist does not exist or is inaccessible: `File '{plist_path}' does not exist or could not be read.`
    - if plistlib fails to load plist: `"{exception_text}: {err}"` where default exception_text is `Unable to load plist`
- There is new Versioner module constant `NO_VERSION_MESSAGE` which is used in both Versioner class and unittests. (8687925)
- Unittest class `TestVersioner` (69d36ca)

Unittest output:

```
test_find_cfbundle_short_version (tests.test_versioner.TestVersioner)
The processor should find version in default CFBundleShortVersionString. ... ok
--
test_find_custom_version (tests.test_versioner.TestVersioner)
The processor should find version under key specified by plist_version_key. ... ok
--
test_no_fail_if_dmg (tests.test_versioner.TestVersioner)
The processor should not raise any exceptions when plist is in the dmg image. ... ok
--
test_no_fail_if_good_env (tests.test_versioner.TestVersioner)
--
test_no_version_found (tests.test_versioner.TestVersioner)
The processor should not find version if plist misses it. ... ok
```

### DIFF
--- a/Code/autopkglib/__init__.py
+++ b/Code/autopkglib/__init__.py
@@ -626,6 +626,18 @@ class Processor:
         else:
             sys.exit(0)
 
+    def load_plist_from_file(plist_path, exception_text="Unable to load plist"):
+        """Load plist from file and return content as dictionary"""
+        if not os.path.exists(plist_path):
+            raise ProcessorError(
+                f"File '{plist_path}' does not exist or could not be read."
+            )
+        try:
+            with open(plist_path, "rb") as f:
+                return plistlib.load(f)
+        except Exception as err:
+            raise ProcessorError(f"{exception_text}: {err}")
+
 
 # AutoPackager class defintion
 
--- a/Code/autopkglib/__init__.py
+++ b/Code/autopkglib/__init__.py
@@ -628,10 +628,6 @@ class Processor:
 
     def load_plist_from_file(plist_path, exception_text="Unable to load plist"):
         """Load plist from file and return content as dictionary"""
-        if not os.path.exists(plist_path):
-            raise ProcessorError(
-                f"File '{plist_path}' does not exist or could not be read."
-            )
         try:
             with open(plist_path, "rb") as f:
                 return plistlib.load(f)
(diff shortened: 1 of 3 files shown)


Code/autopkglib/__init__.py
```diff
@@ -512,6 +512,18 @@ def execute_shell(self):
         else:
             sys.exit(0)
 
+    def load_plist_from_file(plist_path, exception_text="Unable to load plist"):
+        """Load plist from file and return content as dictionary"""
+        if not os.path.exists(plist_path):
+            raise ProcessorError(
+                f"File '{plist_path}' does not exist or could not be read."
+            )
```
nmcspadden: I know you copied this from what was already present, but what's the point of doing this when you're already trying it and catching exceptions below? If we want different output for different possible scenarios, let's catch the specific exception types. Or just stick with the generic one you have, which I think is sufficient for now.

MichalMMac:
@nmcspadden Btw test methods in this PR are using generic "run" methods in the same way I changed it in #601 Since you don't agree with that change #601 should be resolved first so I can implement possible changes here as well before merging.

Code/autopkglib/__init__.py
MichalMMac: You're right. See 7d9d2e0

MichalMMac:
Since #601 was changed I need to update this PR to match the style of the unitests.

MichalMMac:
Rebased against `dev` branch. Mock code duplicated to each test method. Please see 9b40b8e
(no longer Work In Progress).

nmcspadden:
Sorry I dropped the ball on this. Please rebase on master, and this should be good for a merge.

MichalMMac:
Sure. I hope I get to do it soon :-)

MichalMMac:
@nmcspadden Rebased on master branch. Master is now also the merge target. Should I target `dev` branch instead?

nmcspadden:
Switch this to dev and I'll merge it, and then we can rebase #658  and merge it.

MichalMMac:
Rebased on dev a changed target to dev.

PR #601, opened 2019-12-17
Simplify TestURLTextSearcher mock code

MichalMMac:
Small refactoring. Patching only methods where it's necessary.

### DIFF
--- a/Code/tests/test_urltextsearcher.py
+++ b/Code/tests/test_urltextsearcher.py
@@ -47,12 +47,14 @@ class TestURLTextSearcher(unittest.TestCase):
     def tearDown(self):
         pass
 
+    @patch("autopkglib.URLTextSearcher.download_with_curl")
     def match_first(self, mock_download, result_output_var_name="match"):
         """Run processor and expect to find a match"""
         mock_download.return_value = self.web_page
         self.processor.main()
         self.assertEqual(self.processor.env[result_output_var_name], self.first_match)
 
+    @patch("autopkglib.URLTextSearcher.download_with_curl")
     def nomatch_exception(self, mock_download):
         """Run processor and expect exception for lack of match to be raised"""
         mock_download.return_value = self.web_page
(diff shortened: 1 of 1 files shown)


Code/tests/test_urltextsearcher.py
```diff
@@ -47,12 +47,14 @@ def setUp(self):
     def tearDown(self):
         pass
 
+    @patch("autopkglib.URLTextSearcher.download_with_curl")
```
nmcspadden: I don't really like this change, tbh, because I think it makes it slightly more opaque. Each test that uses this function no longer has a choice about what the return values or side effects are when patching the function, because it's done for them at load time of the test case.

For example, if I want to test something more specific with download_with_curl giving me specific output, I can't actually use this function anymore, so I have to copy and paste what it does elsewhere - which essentially eliminates the value of having a primitive function that handles the boilerplate for you.

I'd rather keep this one primitive, and pass in the magic mocks as we currently are.

MichalMMac: > Each test that uses this function no longer has a choice about what the return values or side effects are when patching the function

Each test have a choice of **not** using this method at all. I suspect for some future test method this will be the case. `match_first` method is there only to have fewer lines of code since most of the code in the test methods is exactly the same.

> For example, if I want to test something more specific with download_with_curl giving me specific output, I can't actually use this function anymore

`match_first` method had already set `mock_download.return_value` before this PR. I don't see how moving the decorator here does change anything. If I apply your logic `match_first` should not manipulate mocked method output in any way -> Test methods should set it instead.

MichalMMac:
@nmcspadden Btw test methods in this PR are using generic "run" methods in the same way I changed it in #601  Since you don't agree with that change #601 should be resolved first so I can implement possible changes here as well before merging.

Code/tests/test_urltextsearcher.py
nmcspadden: In my opinion, this breaks the somewhat independent model of the tests.

Each test should declare its individual mocks, and if it happens to call any subroutines (such as match_first), that value is passed in on call. 

Your change makes it so that these tests don't actually declare their own mocks, which in my opinion makes the purpose of the test slightly less clear. I recognize that it's not a functional change regardless, and that in general Python operation this would be the cleaner approach - but I think unit tests should remain as isolated as possible.

If we someday massively change howe URLTextSearcher works, it will take more work to unpack these unit tests later if they depend on other functions that define their own mocks instead.

nmcspadden:
I don't intend to accept this PR, although I do appreciate your work and I don't disagree with your intent.

MichalMMac:
Ok. No problem.

PR #605, opened 2020-01-12
Handle missing url environment variable

homebysix:
There may not be a `url` environment variable in all cases, and this change allows parsing headers regardless.


nmcspadden:
I can't really imagine a scenario in which you'd parse curl headers without having a URL used in the environment, but this is safe enough.

PR #613, opened 2020-02-02
Remove orphaned valid_plist_with_keys() function

homebysix:
This function appears to have been orphaned at some point in the past (probably here: https://github.com/autopkg/autopkg/commit/b754923952d5c3e6eead1e09d5fa4bc7ff193863). A search through the source code reveals that the only reference to the function is in its own return value. If it was ever used again, it would result in a loop. It can be safely removed.

```
% grep -R 'valid_plist_with_keys(' *
Code/autopkg:def valid_plist_with_keys(filename, keys_to_verify):
Code/autopkg:    return valid_plist_with_keys(recipe_plist, keys_to_verify)
```

### DIFF
--- a/Code/autopkg
+++ b/Code/autopkg
@@ -140,13 +140,6 @@ def valid_recipe_plist_with_keys(recipe_plist, keys_to_verify):
     return False
 
 
-def valid_plist_with_keys(filename, keys_to_verify):
-    """Attempts to read a plist file and ensures the keys in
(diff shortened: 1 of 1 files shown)


nmcspadden:
Interesting, looks like it was added back in 2014 in https://github.com/autopkg/autopkg/commit/b754923952d5c3e6eead1e09d5fa4bc7ff193863. It was orphaned some time ago, it seems, but looks like it was intended to make listing recipes more efficient.

PR #626, opened 2020-03-26
Use a GET request in URLDownloader.prefetch_filename()

UniversalSuperBox:
Some servers disallow HTTP HEAD requests, used to get information about a file. This is the default request type that curl sends when '--head' is passed.

By passing '--request GET' to curl, this default behavior is overridden. curl will now perform a GET request but stop after it receives the headers. This ensures that we get the correct headers for every file (some servers may return different headers for GET vs HEAD) and don't receive HTTP Forbidden errors for an incorrect request type.

Fixes https://github.com/autopkg/autopkg/issues/624

Does anyone have some recipes that should be tested with this change? It only affects prefetch_filename.

### DIFF
--- a/Code/autopkglib/URLDownloader.py
+++ b/Code/autopkglib/URLDownloader.py
@@ -170,7 +170,7 @@ class URLDownloader(URLGetter):
     def prefetch_filename(self):
         """Attempt to find filename in HTTP headers."""
         curl_cmd = self.prepare_base_curl_cmd()
-        curl_cmd.extend(["--head"])
+        curl_cmd.extend(["--head", "--request", "GET"])
 
         raw_headers = self.download_with_curl(curl_cmd)
         header = self.parse_headers(raw_headers)


MichalMMac:
I am all for it if @UniversalSuperBox did test this.

UniversalSuperBox:
I've tested this change with the recipes on GitHub I could find which used `prefetch_filename`:

https://github.com/overhacked/autopkg-recipes/blob/250cbc7df315a89ff986edafedf4aa4fb5037644/Nextcloud/Nextcloud.download.recipe
https://github.com/aarondavidpolley/Autopkg-Recipes/blob/master/Vivi/Vivi.download.recipe
https://github.com/autopkg/jazzace-recipes/blob/04dbfbba593ae60e91d2ff53b417f108e6f30c15/Nemetschek/Vectorworks.download.recipe

In each case, autopkg found the same filename as it did without this change.

The test case I was fixing against is for adoptopenjdk: https://github.com/UniversalSuperBox/autopkg_recipes/blob/7278c0f92a72917869d4386a3402202c85e69416/AdoptOpenJDK8/openjdk8-binaries.download.recipe

This recipe does not work without this change, and it works correctly with it.

nmcspadden:
Thank you for the testing!

PR #648, opened 2020-07-01
Windows: Base Compatibility

linuxfood:
This PR introduces basic scaffolding and workflow changes for Windows support:
* Improves testing workflow. This portion, the first 4 commits, can easily be moved to a separate PR as needed.
* Ensures that text files have consistent line endings, makes `pip install -r requirements.txt` work on windows (and linux, technically), as well as fixing a couple tooling issues around Python versioning that caused me no end of frustration figuring out.
* Makes `requirements.txt` windows compatible.
* Makes all the existing tests pass cleanly on Windows (or be skipped). New platform independent functionality is added for discovering curl and git.

Test results from top of stack [Add platform independent way to find a utility](https://github.com/autopkg/autopkg/pull/648/commits/6b71a296384bc43a9d4bb8fb1435e183da7d8d40) all look good:
* [Autopkg macOS unittests](https://gist.github.com/linuxfood/58bf48a8a5b74d9242964215f1eee455)
* [Autopkg info -p output](https://gist.github.com/linuxfood/e87a6857bc308242457e1e13ee324dba)
* [Autopkg run -vv output](https://gist.github.com/linuxfood/47246fcae22586a2cc4d6a1e1d4ade49)

PR #651, opened 2020-07-07
Windows: Add signtool processor 

linuxfood:
Add a new processor for Windows `signtool.exe`. Additionally, this includes a wrapper library around `xattr` to so that `URLDownloader` has no more platform specific code in it. This PR depends on #648.

The signtool integration/functional tests pass on Windows and are correctly excluded on other platforms. With the inclusion of the `xattr` changes, autopkg is also now able to download a package on Windows.

Test results:
* [Autopkg Windows example output](https://gist.github.com/linuxfood/eb4efe457bb2415b4c5a9acab52d3b61) - showing successfully downloading a package on Windows. The CodeSignatureVerifier error is expected and a future PR will make sure that CodeSignatureVerifier explicitly to run on Windows.
* [Autopkg macOS run -vv output](https://gist.github.com/linuxfood/f04bd008b5e18d992d1ca94bfb39e4eb) - demonstrating that macOS still works after these changes.

### DIFF
--- a/Code/autopkglib/SignToolVerifier.py
+++ b/Code/autopkglib/SignToolVerifier.py
@@ -1,0 +1,149 @@
+#!/usr/local/autopkg/python
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""See docstring for SignToolVerifier class"""
+
+import os
+import os.path
+import subprocess
+from typing import Any, Dict, List, Optional
+
+from autopkglib import Processor, ProcessorError
+
+__all__ = ["SignToolVerifier"]
+
+
+def signtool_default_path() -> Optional[str]:
+    """Looks for signtool in a few well known paths. Deliberately naive."""
+    for program_files_candidate, arch in (
+        (os.environ.get("ProgramFiles(x86)"), "x64"),
+        (os.environ.get("ProgramFiles(x86)"), "x86"),
+        (os.environ.get("ProgramFiles"), "x64"),
+        (os.environ.get("ProgramFiles"), "x86"),
+    ):
+        if program_files_candidate is None:
+            continue
+        candpath = os.path.abspath(
+            os.path.join(
+                program_files_candidate, r"Windows Kits\10\bin", arch, "signtool.exe"
+            )
+        )
+        if os.path.exists(candpath):
+            return candpath
+    return None
+
+
+class SignToolVerifier(Processor):
+    """Verifies an authenticode signed installer using the Microsoft SDK
+    signtool executable."""
+
+    EXTENSIONS: List[str] = [".exe", ".msi"]
+
+    # TODO: How much of this is needed to act as a drop-in replacement in an
+    # override recipe??
+    input_variables: Dict[str, Any] = {
+        "DISABLE_CODE_SIGNATURE_VERIFICATION": {
+            "required": False,
+            "description": ("Prevents this processor from running."),
+        },
+        "input_path": {
+            "required": True,
+            "description": (
+                "File path to an `.msi` or `.exe` file for Authenticode verification",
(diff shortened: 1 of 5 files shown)


Code/autopkglib/SignToolVerifier.py
```diff
@@ -0,0 +1,131 @@
...
+        signtool_path: Optional[str] = None,
+    ) -> bool:
+        """
+        Runs 'signtool.exe /pa <path>'. Returns True if signtool exited with 0
+        and False otherwise.
+        """
+        if not additional_arguments:
+            additional_arguments = []
+
+        signtool_bin = signtool_path if signtool_path else "signtool.exe"
```
nmcspadden, line 78: I'd greatly prefer an absolute path here. I know it's not very Windows-like, but relative paths keep me up at night.

Code/autopkglib/SignToolVerifier.py
```diff
@@ -0,0 +1,131 @@
...
+
+        return proc.returncode == 0
+
+    def main(self):
+        if self.env.get("DISABLE_CODE_SIGNATURE_VERIFICATION"):
+            self.output("Authenticode verification disabled for this recipe run.")
+            return
+
+        input_path = self.env["input_path"]
+        signtool_path = self.env["signtool_path"]
```
nmcspadden, line 119: This isn't a required variable, so this could fail if undefined. Use `get()` here instead.

Code/autopkglib/SignToolVerifier.py
```diff
@@ -0,0 +1,131 @@
...
+        additional_arguments = self.env["additional_arguments"]
```
nmcspadden, line 120: Same here.

Code/autopkglib/SignToolVerifier.py
linuxfood: Oh, thanks for calling this out. This does need to change. What would you think of getting rid of this fallback entirely? I'm thinking that since `signtool_path` has an absolute path as its default value in the input variables, it should be fine to have the only two cases here be: using the default, or manually specifying the variable. In either case, if the binary doesn't exist, the processor should error out explicitly.

Code/autopkglib/SignToolVerifier.py
linuxfood: Hm. Can you clarify the issue? Here's what I understand: 
* `input_path` is a required input variable. If it's unspecified the base class method `Processor.process` will fail with an error.
* `signtool_path` is optional, but the input variables specification provides a default. If no input for it is given `Processor.process` will fill in the default value. So, for this case, it seems like the only way for this be non-existent is if thing goes out of its way to bypass `Processor.process`.

Code/autopkglib/SignToolVerifier.py
linuxfood: `additional_arguments` has a default value specified as well. Thoughts?

Code/autopkglib/SignToolVerifier.py
```diff
@@ -0,0 +1,131 @@
...
+                "File path to an `.msi` or `.exe` file for Authenticode verification",
+            ),
+        },
+        "signtool_path": {
+            "required": False,
+            "description": (
+                "The path to signtool.exe. This varies between versions of the "
+                "Windows SDK, so you can explicitly set that here in an override."
+            ),
+            "default": r"C:\Program Files (x86)\Windows Kits\10\bin\x64\signtool.exe",
```
nmcspadden, line 50: If we're going to eliminate the fallback, this needs to use `os.path.join(os.getenv['PROGRAMFILES']` instead of hardcoding C:.

Code/autopkglib/SignToolVerifier.py
nmcspadden: I'm okay with just eliminating it entirely, if it's defined in the input with a valid default. I made a comment about this specifically - if we're going to provide the path in the default value for the input variable, let's not hardcode C:.

Code/autopkglib/SignToolVerifier.py
nmcspadden: Sorry, I managed to miss the default values in the input variables.

Code/autopkglib/SignToolVerifier.py
nmcspadden: I managed to miss that, sorry.

Code/autopkglib/SignToolVerifier.py
linuxfood: Sounds good to me. I'll update the PR today.

linuxfood: Done

PR #656, opened 2020-07-22
Windows: Fix inputs bug in SigntoolProcessor

linuxfood:
The `signtool_path` input variable had a default value as well as
setting `required: True`. Consequently, it didn't work when running as
part of a larger recipe due to inconsistencies in how autopkg interprets
variables.


jgstew:
Seems like this means that if an input variable has a default then `required` should always be `false`

Might be useful to expand this: https://github.com/autopkg/autopkg/wiki/Developing-Custom-Processors

I've been developing custom processors lately and it is hard to tell sometimes what the right way to go about it is.

PR #658, opened 2020-07-30
Windows/autopkg recipe support

linuxfood:
Add capabilities to read plist files from inside zip archives in addition to dmg and iso images.  Add unit tests as well. There is likely some merging with the work done in #600 that will need to happen either in this PR or in a merge with master and dev.

### DIFF
--- a/Code/autopkglib/Unarchiver.py
+++ b/Code/autopkglib/Unarchiver.py
@@ -18,8 +18,11 @@
 import os
 import shutil
 import subprocess
+import tarfile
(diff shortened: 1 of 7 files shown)


nmcspadden:
I think it makes sense to merge in #600 into dev, and rebase yours on top.

nmcspadden:
I've merged #600 so this will need to be rebased to work out the conflicts.

linuxfood:
I've rebased on top of #600.

PR #660, opened 2020-07-31
Support exporting processors outside of AutoPkg

tboyko:
Currently, if a processor, such as `URLGetter` is imported into an external codebase (for example: `from autopkglib import URLGetter`), `self.env` is set to `None`. This causes problems as `self.env` is expected to be, at a minimum, an empty dictionary.

This commit simply sets `self.env` to `{}` if not provided otherwise.

Here is an example of the error seen before the change:

```
  File "/Library/AutoPkg/autopkglib/__init__.py", line 661, in process
  File "/Library/AutoPkg/autopkglib/__init__.py", line 468, in process
    self.main()
  File "/Library/AutoPkg/autopkglib/MunkiImporter.py", line 250, in main
    matchingitem = library.find_matching_pkginfo(pkginfo)
  File "/Library/AutoPkg/autopkglib/munkirepolibs/MunkiLibAdapter.py", line 39, in find_matching_pkginfo
    match = self.munkiimportlib.find_matching_pkginfo(self.repo, pkginfo)
  File "/usr/local/munki/munkilib/admin/munkiimportlib.py", line 234, in find_matching_pkginfo
    catdb = make_catalog_db(repo)
  File "/usr/local/munki/munkilib/admin/munkiimportlib.py", line 137, in make_catalog_db
    plist = repo.get('catalogs/all')
  File "/usr/local/munki/munkilib/munkirepo/SimpleMDMRepo.py", line 82, in get
    resp = self._curl(resource_identifier)
  File "/usr/local/munki/munkilib/munkirepo/SimpleMDMRepo.py", line 46, in _curl
    curl_cmd = self.getter.prepare_curl_cmd()
  File "/Library/AutoPkg/autopkglib/URLGetter.py", line 68, in prepare_curl_cmd
    return [self.curl_binary(), "--compressed", "--location"]
  File "/Library/AutoPkg/autopkglib/URLGetter.py", line 41, in curl_binary
    if "CURL_PATH" in self.env and is_executable(self.env["CURL_PATH"]):
TypeError: argument of type 'NoneType' is not iterable
  File "/Library/AutoPkg/autopkglib/__init__.py", line 661, in process
argument of type 'NoneType' is not iterable
Failed.
```

Does this change seem appropriate?

### DIFF
--- a/Code/autopkglib/URLGetter.py
+++ b/Code/autopkglib/URLGetter.py
@@ -29,6 +29,11 @@ class URLGetter(Processor):
 
     description = __doc__
 
+    def __init__(self, env=None, infile=None, outfile=None):
+        super().__init__(env, infile, outfile)
+        if not self.env:
+            self.env = {}
+
     def curl_binary(self):
         """Return a path to a curl binary, priority in the order below.
         Return None if none found.
--- a/Code/autopkglib/__init__.py
+++ b/Code/autopkglib/__init__.py
@@ -378,7 +378,7 @@ class Processor:
     returns a new or updated property list that can be processed further.
     """
 
-    def __init__(self, env={}, infile=None, outfile=None):
+    def __init__(self, env=None, infile=None, outfile=None):
         # super(Processor, self).__init__()
         self.env = env
         if infile is None:
--- a/Code/autopkglib/__init__.py
+++ b/Code/autopkglib/__init__.py
@@ -378,7 +378,7 @@ class Processor:
     returns a new or updated property list that can be processed further.
     """
 
-    def __init__(self, env=None, infile=None, outfile=None):
+    def __init__(self, env={}, infile=None, outfile=None):
         # super(Processor, self).__init__()
         self.env = env
         if infile is None:


tboyko:
On a related note, I thought that `URLGetter(env={})` would also solve the issue (per https://github.com/autopkg/autopkg/blob/master/Code/autopkglib/__init__.py#L381) , however it doesn't seem to make a difference.

nmcspadden:
Unfortunately, this solution isn't workable. It works, but for the wrong reasons. This blog post explains it much better than I can:
https://florimond.dev/blog/articles/2018/08/python-mutable-defaults-are-the-source-of-all-evil/

The tl;dr is that if you pass a mutable data structure as a default argument (such as {} or []), that data structure persists for _every single future call of that argument_. In AutoPkg's case, this would mean that every processor that gets initialized would share the same starting state for the env, rather than inheriting it from the recipe's runtime. This could lead to all sorts of unexpected behavior that would be extremely difficult to troubleshoot and diagnose, especially from the somewhat limited AutoPkg output.

What might be better is what I suggested in Slack, putting this code block at the top of the `main()` function in URLGetter itself:
```
if not self.env:
  self.env = {}
```

or creating an `__init__()` function and calling its super to get both worlds to work.

tboyko:
Made changes per your recommendations. How does this look?

nmcspadden:
That's a big improvement. Now the usual request:
Please show output from an existing recipe that relies on a URLGetter-based processor to show that it doesn't affect existing behavior

tboyko:
This example uses URLDownloader, which inherits from URLGetter, and so I believe it is sufficient. Let me know if otherwise.

```bash
$ autopkg run -v GoogleChrome.munki -k MUNKI_REPO_PLUGIN="SimpleMDMRepo" -k extract_icon=True
Processing GoogleChrome.munki...
WARNING: GoogleChrome.munki is missing trust info and FAIL_RECIPES_WITHOUT_TRUST_INFO is not set. Proceeding...
URLDownloader
URLDownloader: Storing new Last-Modified header: Fri, 24 Jul 2020 21:14:19 GMT
URLDownloader: Storing new ETag header: "65cb9b"
URLDownloader: Downloaded /Users/example/Library/AutoPkg/Cache/com.github.autopkg.munki.google-chrome/downloads/GoogleChrome.dmg
EndOfCheckPhase
CodeSignatureVerifier
CodeSignatureVerifier: Mounted disk image /Users/example/Library/AutoPkg/Cache/com.github.autopkg.munki.google-chrome/downloads/GoogleChrome.dmg
CodeSignatureVerifier: Verifying code signature...
CodeSignatureVerifier: Deep verification enabled...
CodeSignatureVerifier: Strict verification disabled...
CodeSignatureVerifier: /private/tmp/dmg.iRNHR1/Google Chrome.app: valid on disk
CodeSignatureVerifier: /private/tmp/dmg.iRNHR1/Google Chrome.app: satisfies its Designated Requirement
CodeSignatureVerifier: /private/tmp/dmg.iRNHR1/Google Chrome.app: explicit requirement satisfied
CodeSignatureVerifier: Signature is valid
MunkiImporter
NOTICE: The SimpleMDMRepo plugin ignores the MUNKI_REPO value. "[redacted]" will be used instead.
MunkiImporter: Using repo lib: MunkiLibAdapter
MunkiImporter:         plugin: SimpleMDMRepo
MunkiImporter:           repo: /Volumes/html
MunkiImporter: Copied pkginfo to: /Volumes/html/pkgsinfo/apps/GoogleChrome-84.0.4147.105
MunkiImporter:            pkg to: /Volumes/html/pkgs/apps/GoogleChrome-84.0.4147.105.dmg
MunkiImporter:           icon to: /Volumes/html/icons/GoogleChrome.png
Receipt written to /Users/example/Library/AutoPkg/Cache/com.github.autopkg.munki.google-chrome/receipts/GoogleChrome-receipt-20200804-151813.plist

    Download Path                                                                                               
    -------------                                                                                               
    /Users/example/Library/AutoPkg/Cache/com.github.autopkg.munki.google-chrome/downloads/GoogleChrome.dmg  

The following new items were imported into Munki:
    Name          Version        Catalogs  Pkginfo Path                     Pkg Repo Path                        Icon Repo Path    
    ----          -------        --------  ------------                     -------------                        --------------    
    GoogleChrome  84.0.4147.105  testing   apps/GoogleChrome-84.0.4147.105  apps/GoogleChrome-84.0.4147.105.dmg  GoogleChrome.png  
```

PR #666, opened 2020-08-06
Bugfix: Skip chocolatey tests on macOS

linuxfood:
Python `subprocess` raises `FileNotFound` which wasn't being caught, thus tests were failing on macOS when a chocolatey test case tried to skip based on running `choco.exe --version`.

### DIFF
--- a/Code/tests/test_chocolatey_packager.py
+++ b/Code/tests/test_chocolatey_packager.py
@@ -37,10 +37,21 @@ def get_mocked_writes(mock: unittest.mock.MagicMock) -> str:
     return res
 
 
+def check_for_choco() -> bool:
+    try:
+        return (
+            subprocess.run(
+                [r"C:\ProgramData\chocolatey\bin\choco.exe", "--version"]
+            ).returncode
+            == 0
+        )
+    except FileNotFoundError:
+        pass
+    return False
+
+
 @unittest.skipUnless(sys.platform.startswith("win"), "requires Windows")
-@unittest.skipUnless(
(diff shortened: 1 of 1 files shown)


Code/autopkglib/github/__init__.py
```diff
@@ -20,6 +20,7 @@
 import os
 import re
 import tempfile
+from urllib.parse import quote
```
nmcspadden, line 23: This whole file should disappear in a rebase, because I think this was merged in when you rebased dev onto master.

Code/tests/test_chocolatey_packager.py
```diff
@@ -37,10 +37,16 @@ def get_mocked_writes(mock: unittest.mock.MagicMock) -> str:
     return res
 
 
+def check_for_choco() -> bool:
+    try:
+        return subprocess.run(["choco.exe", "--version"]).returncode == 0
```
nmcspadden, line 42: This is generally going to be fine, but I greatly prefer using an absolute path. Choco installs in a fixed path (I think?), and trusting the powershell path makes me itchy.

Code/autopkglib/github/__init__.py
linuxfood: Yes, I just didn't want to make the merge PR muddied with this bugfix PR. I'll pull and repush and everything will look much nicer.

Code/tests/test_chocolatey_packager.py
linuxfood: It does appear to have a fixed installation path. I'll change this.

PR #670, opened 2020-08-25
Fix file-backed preference saving

linuxfood:
When adding global preference support for Windows, saving file-backed (e.g., a json or plist file specified with `--prefs`) became broken on macOS and Linux. This fixes that.

The PR also includes a change that moves the definitions for `log` and `log_err` nearer the top of the module so that they can be used more uniformly, rather than some code paths using `print`.

### DIFF
--- a/Code/autopkglib/__init__.py
+++ b/Code/autopkglib/__init__.py
@@ -58,6 +58,19 @@ def is_linux():
     return "Linux" in platform.platform()
 
 
+def log(msg, error=False):
+    """Message logger, prints to stdout/stderr."""
+    if error:
+        print(msg, file=sys.stderr)
+    else:
+        print(msg)
+
+
+def log_err(msg):
+    """Message logger for errors."""
+    log(msg, error=True)
+
+
 try:
     from CoreFoundation import (
         CFPreferencesAppSynchronize,
@@ -299,19 +312,6 @@ def get_all_prefs():
     """Return a dict (or an empty dict) with the contents of all
     preferences in the domain."""
     return globalPreferences.get_all_prefs()
-
-
-def log(msg, error=False):
-    """Message logger, prints to stdout/stderr."""
-    if error:
-        print(msg, file=sys.stderr)
-    else:
-        print(msg)
-
-
-def log_err(msg):
-    """Message logger for errors."""
-    log(msg, error=True)
 
 
 def get_identifier(recipe):
--- a/Code/autopkglib/__init__.py
+++ b/Code/autopkglib/__init__.py
@@ -133,16 +133,18 @@ class Preferences:
 
     def __init__(self):
         """Init."""
-        self.prefs = {}
+        self.prefs: VarDict = {}
         # What type of preferences input are we using?
-        self.type = None
+        self.type: Optional[str] = None
         # Path to the preferences file we were given
-        self.file_path = None
+        self.file_path: Optional[str] = None
         # If we're on macOS, read in the preference domain first.
         if is_mac():
             self.prefs = self._get_macos_prefs()
-        elif is_windows():
-            self.prefs = self._get_windows_prefs()
+        else:
+            self.prefs = self._get_file_prefs()
+        if not self.prefs:
+            log_err("WARNING: Did not load any default preferences.")
 
     def _parse_json_or_plist_file(self, file_path):
         """Parse the file. Start with plist, then JSON."""
@@ -210,9 +212,10 @@ class Preferences:
                     prefs[key] = self._get_macos_pref(key)
         return prefs
 
-    def _get_windows_prefs(self):
+    def _get_file_prefs(self):
         r"""Lookup preferences for Windows in a standardized path, such as:
-        `C:\\Users\username\AppData\Local\Autopkg\config.{plist,json}`
+        * `C:\\Users\username\AppData\Local\Autopkg\config.{plist,json}`
+        * `/home/username/.config/Autopkg/config.{plist,json}`
         Tries to find `config.plist`, then `config.json`."""
 
         config_dir = appdirs.user_config_dir(APP_NAME, appauthor=False)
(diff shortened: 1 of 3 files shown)


Code/autopkglib/__init__.py
```diff
@@ -277,8 +293,10 @@ def set_pref(self, key, value):
         # On macOS, write it back to preferences domain if we didn't use a file
         if is_mac() and self.type is None:
             self._set_macos_pref(key, value)
-        elif is_windows() and self.file_path is not None:
+        elif self.file_path is not None:
             self.write_file()
+        else:
+            log(f"WARNING: Preference change {key}=''{value}'' was not saved.")
```
nmcspadden, line 299: Should this be log_err() instead?

linuxfood: I'm not sure. It could be. This case will happen whenever you're running autopkg on Linux or Windows, the default config (such as `/home/username/.config/Autopkg/config.json`) doesn't exist and you haven't used the `--prefs` option. I don't believe it's possible for it to happen on mac, since either the preference domain will be used, or `--prefs` will have been passed to autopkg.

Thoughts?

nmcspadden: Even if this is not something that should happen commonly, this still suggests to me that it's an error condition. It even has "WARNING" in the text right there. If it's an error condition of any kind, I kinda feel like like it should be `log_err` and go to stderr.

I'm not super strongly focused on this, but we should be consistent. If there's enough precedence for other warnings going to stdout and not stderr, I'll drop it.

linuxfood: Changed this to `log_err`. I agree about the `stdout` / `stderr` division.

linuxfood:
I've also merged in master to dev and rebased this PR on top of that.

.isort.cfg
```diff
@@ -4,4 +4,4 @@ include_trailing_comma=True
 force_grid_wrap=0
 use_parentheses=True
 line_length=88
-known_third_party = appdirs,certifi,xattr
+known_third_party = appdirs,certifi,grp,pwd,xattr
```
nmcspadden, line 7: How did these end up in here?

CHANGELOG.md
```diff
@@ -1,4 +1,24 @@
-### [2.1.1](https://github.com/autopkg/autopkg/compare/v2.1...HEAD) (Unreleased)
```
nmcspadden, line 1: Same here - can we just remove this from the PR?

Code/autopkglib/version.plist
```diff
@@ -3,6 +3,6 @@
 <plist version="1.0">
 <dict>
 	<key>Version</key>
-	<string>2.1.1</string>
```
nmcspadden, line 6: And this too.

CHANGELOG.md
linuxfood: Sure, I'll undo the master merge.

.isort.cfg
linuxfood: seed-isort is a capricious god?

linuxfood: TIL- `seed-isort-config` is no longer necessary for `isort>=5.0`, which we're now using.

CHANGELOG.md
linuxfood: Actually, how about I setup a separate dev-master merge PR and then once that gets pulled in, then this PR won't include the changelog, etc change.

linuxfood:
Okay. This now only contains the preferences bugfix, and does not contain changes to `.isort.cfg`: https://github.com/autopkg/autopkg/pull/673 should improve spurious additions+removals, as well.

Code/tests/test_autopkglib.py
```diff
@@ -256,88 +255,6 @@ def test_get_identifier_from_recipe_file_returns_none(self, mock_load, mock_read
...
-    @patch("autopkglib.CFPreferencesSetAppValue")
-    def test_set_pref_mac(self, mock_sav, mock_ismac):
-        """set_pref should change the prefs object."""
-        mock_ismac.return_value = True
-        fake_prefs = autopkglib.Preferences()
-        fake_prefs.set_pref("TEST_KEY", "fake_value")
-        value = fake_prefs.get_pref("TEST_KEY")
-        self.assertEqual(value, "fake_value")
-        mock_sav.assert_called()
-
```
linuxfood, line 340: All these tests were moved into the `test_preferences` module introduced below. Some minor refactoring was done as well.

Code/autopkglib/__init__.py
```diff
@@ -248,6 +265,7 @@ def _write_json_file(self):
     def _write_plist_file(self):
         """Write out the prefs into a Plist."""
         try:
+            assert self.file_path is not None
```
nmcspadden, line 268: Do you want to leave these assertions in? I feel like this should be surfaced as an exception with a human-readable error message rather than an assert failure traceback.

linuxfood: I think they should be left in. Any instance of this assertion tripping is a bug in autopkg, not anything that should be possible during normal usage. You'll note that both `_write_plist_file` and `_write_json_file` are only called through `write_file` which returns if `self.file_path` is `None`. So, the only way this happens if someone calls one of these directly (which nothing does), or the code is changed in such a way that validation goes missing. I'll add an explanation here to the assert like `assert self.file_path is not None, "Internal logic error in preference saving. Please report this to autopkg developers."` to add a little clarity. Though, it's worth noting that this won't cause autopkg to crash, since `AssertionError` derives from `Exception` which both functions swallow anyway...

Code/autopkglib/version.plist
Tinellmann1979: > And this too.

Tinellmann1979 (commented):
With cared use the app would be better

Tinellmann1979 (commented):
Needs more

PR #697, opened 2020-11-28
Remove path component confusion in munkiimporter

octomike:
So we moved our munki repo recently and from that point on autopkg imported packages with broken `installer_item_location` values. It turns out this was because our mount path contained the substring `pkgs/`

https://github.com/autopkg/autopkg/blob/af858c033643089b8ecbccf0a8ad691d4c28a289/Code/autopkglib/MunkiImporter.py#L384

```
>>> '/mnt/munkipkgs/pkgs/apps/Keka/Keka-1.2.4.dmg'.partition('pkgs/')
('/mnt/munki', 'pkgs/', 'pkgs/apps/Keka/Keka-1.2.4.dmg')
```

This PR *slightly* improves the handling of that by adding a leading `/` to the split value. Conflicts may still occur though.

### DIFF
--- a/Code/autopkglib/MunkiImporter.py
+++ b/Code/autopkglib/MunkiImporter.py
@@ -381,7 +381,8 @@ class MunkiImporter(Processor):
 
         # import pkg
         install_path = library.copy_pkg_to_repo(pkginfo, self.env["pkg_path"])
-        pkginfo["installer_item_location"] = install_path.partition("pkgs/")[2]
+        install_prefix = os.path.join(library.munki_repo, 'pkgs')
+        pkginfo["installer_item_location"] = os.path.relpath(install_path, install_prefix)
         self.env["pkg_repo_path"] = install_path
 
         if self.env.get("uninstaller_pkg_path"):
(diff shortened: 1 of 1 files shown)


nmcspadden:
I'm not sure I understand exactly what problem this is solving. It seems like changing your mount path would be the obvious solution here that doesn't require changing AutoPkg.

octomike:
Yes, that would be one solution, now that I know what the problem is.
My thought was just that making this slightly more robust might stop other people falling into this trap.

gregneagle:
I see the problem, but if we’re going to fix it, let’s fix it correctly, which would be to first remove the mount path from the start of the path, and _then_ look for the pkgs directory.

octomike:
Totally agree. I updated the feature branch in the least intrusive way I could come up with. The icon thing is a bit ugly, because the path can be also an empty string.

octomike:
> This seems reasonable now, but please change this over to the dev branch.

No problem. Rebased, force pushed, changed target to `dev`.

Issue #276, opened 2016-04-29
Not properly handling SSH git URLs


jessepeterson:
If you try add a repo using autopkg like this:

``` bash
Code/autopkg repo-add user@host.example.com:repo.git
```

You get attempts to clone a GitHub URL:

``` bash
Code/autopkg repo-add user@host.example.com:repo.git
Attempting git clone...
ERROR: Cloning into '/Users/jesse/Library/AutoPkg/RecipeRepos/com.github.autopkg.user@host.example.com:repo'...
fatal: repository 'https://github.com/autopkg/user@host.example.com:repo.git/' not found
```

I believe this is due to this code:
https://github.com/autopkg/autopkg/blob/master/Code/autopkg#L631

Which seems to assume you're using a Github URL if `urlparse()` doesn't successfully return a URL scheme.

Would looking for an `@` and simply returning the URL be adequate here? Or maybe replacing the re `r"\w+"` with `r"^[\w-]+$"` instead (which works), though now you have directories like this:

```
Code/autopkg repo-add user@host.example.com:repo.git
Attempting git clone...

Adding /Users/jesse/Library/AutoPkg/RecipeRepos/user@host.example.com:repo to RECIPE_SEARCH_DIRS...
Updated search path:
[...]
```

Happy to submit a PR if a direction is decided on.

gkluoe:
I'd really like to see this implemented - could it be reopened? Happy to submit a PR if @jessepeterson doesn't need this any more.

nmcspadden:
If you wish to work on it, PRs will certainly be considered. This would be at the bottom of my personal priority list, so I wouldn't be getting to it any time soon.

choules:
Since we ran into this issue yesterday, I have submitted the small pull request #744 to fix this.

PR #744, opened 2021-05-18
Fix handling of scp syntax for git urls

choules:
These changes fix some problems already mentioned in Issue #276. Using the common scp syntax for git urls was not properly processed.

We convert those kind of urls to a proper ssh url, so that

`autopkg repo-add git@privategitserver:namespace/repo.git`

is actually equivalent to

`autopkg repo-add ssh://git@privategitserver/namespace/repo.git`

Furthermore we are removing the username – if present – from the url, to keep the directory path free from (probably unwanted) "@" characters.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-### [2.3](https://github.com/autopkg/autopkg/compare/v2.2...HEAD) (Unreleased)
+### [2.3](https://github.com/autopkg/autopkg/compare/v2.2...v2.3) (March 01, 2021)
 
 NEW FEATURES:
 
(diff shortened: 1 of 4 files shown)


nmcspadden:
This is neat, I like it. Can you please show the output of autopkg repo-add using both this syntax, and a regular pattern (such as "recipes" or "nmcspadden-recipes") to show that it works with your changes?

choules:
Nice catch, @nmcspadden. I definitely need to gain more experience with Python. It should work correctly now.

nmcspadden:
Please change the target of this over to the dev branch, thanks. Then I'll merge.

choules:
Done.

PR #753, opened 2021-06-25
Add option in MunkiImporter to set PackageCompleteURL from env["url"]

paulsuh:
This is so that the munki clients will pull packages from the original site, especially useful if your organization has many people who are remote.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ autopkgc
 
 # IDE-specific files
 .vscode
+.idea
--- a/Code/autopkglib/MunkiImporter.py
+++ b/Code/autopkglib/MunkiImporter.py
@@ -129,6 +129,14 @@ class MunkiImporter(Processor):
             ),
             "required": False,
         },
+        "set_PackageCompleteURL": {
+            "description": (
+                "If set to True, the PackageCompleteURL key for the package will be "
+                "set to the url of the package, so that clients will pull from "
+                "the original source rather than the repository. "
+            ),
+            "required": False,
+        },
     }
     output_variables = {
         "pkginfo_repo_path": {
@@ -357,6 +365,11 @@ class MunkiImporter(Processor):
                         )
                     )
                 item["version_comparison_key"] = self.env["version_comparison_key"]
+
+        # set the PackageCompleteURL to the url from env so that clients pull the
+        # package from the original source, not the repo
+        if self.env.get("set_PackageCompleteURL"):
+            pkginfo["PackageCompleteURL"] = self.env["url"]
 
         # check to see if this item is already in the repo
         if self.env.get("force_munkiimport"):
--- a/Code/autopkglib/MunkiImporter.py
+++ b/Code/autopkglib/MunkiImporter.py
@@ -131,9 +131,9 @@ class MunkiImporter(Processor):
         },
         "set_PackageCompleteURL": {
             "description": (
-                "If set to True, the PackageCompleteURL key for the package will be "
+                "If set to True, the PackageCompleteURL key will be "
                 "set to the url of the package, so that clients will pull from "
-                "the original source rather than the repository. "
+                "the original download location rather than the repository. "
             ),
             "required": False,
         },


arubdesu:
I love Paul Suh sightings! 😃 Long time no see! 
There are over a dozen input variables in MunkiImporter, which I think is about the highest of any processor in core, I'm not sure how folks will weigh in about the potential of sprawl adding support difficulty. That being said, there's MunkiImporter work in the (currently) previous open PR (#749) that I'm using a variant of now, I'd love to test this with the MS Office suite but I don't see how this prevents the actual pkg from being stored in the munki repo? I guess it's fine to leave in there since that's not the majority of the bandwidth and overhead.

You may want to squash commits, although GitHub would do that automatically.
I don't think anyone minds you're modifying the gitignore for intelliJ IDE project stuff.

paulsuh:
I'm coming back into this area now -- I've been working on other things for a while. Nice to be back. 

Overall It's a minor change for a specific use case thanks to the pandemic. 

On the one hand yes it's one more in a long list of options for MunkiImporter. On the other hand, what's the big deal with one more option in the list that's already that long? Most people don't use all of the options in any case. 

Yes, the actual pkg gets imported into the repo but that doesn't make a difference for most use cases and it's also useful for a Munki branch that I'm working on. 

I tried (and have a working version of) an approach based on a delegate-style intervention (a class that has willRun<ProcessorName> and didRun< ProcessorName> methods), but that struck me as overkill. Also, the delegate-style intervention requires more knowledge of programming, and more code changes. I figure the pluses and minuses are a wash and it's better to go with less code. I may still submit the delegate code as a separate pull request. 

Yes, I could squash commits but it's only two vs. one, mostly because I forgot to set up the pre-commit tests in my fresh development venv and wanted another commit to make sure everything passed.

gregneagle:
A side-note not directly reflecting on the content of this PR: setting PackageCompleteURL to the vendor's URL seems incredibly fragile. 
If the URL is a "stable" URL ("https://foo.com/downloads/bar-latest") then on the next update, the package checksum will be wrong and you'll have lots of complaining clients. 
If the vendor has unique URLs for each version, they may make older versions unavailable and again you'll have lots of complaining clients, if any of them are still using the older version (or you haven't promoted the new version to production just yet).

gregneagle:
I wonder if this can be done with no code changes to MunkiImporter by using PkgInfoMerger? In fact, this might be able to be an override for many recipes.
Also: MunkiImporter supports `additional_makepkginfo_options`, which could also perhaps handle this use-case...

arubdesu:
MS Office in particular does provide 'stable'/permalink-y CDN URL's for both latest and each distinct version, but Greg's point is important that you'll have hash mis-matches if your updates to pkginfo don't make it out the door fast enough to beat the previous version in the cached catalogs on clients... or maybe you just temporarily deny service 😅 and mute the hash mis-match errors during the window it's out-of-sync in?

paulsuh:
The original package still gets added to the repo so you can always edit the pkgsinfo to point to that if you start getting errors. In most use cases I think that the vast majority of the load will come while the primary outside URL is live, which could be a big help for places that don't have a lot of upload bandwidth and have many remote machines. 

I have some modifications I'm working on in Munki to add a capability similar to Jamf's network segments. I may add an option to fall back to the 'standard' location URL if the PackageCompleteURL gives an error.

paulsuh:
I checked and using `additional_makepkginfo_options` won't work -- that's for additional command line args to the `makepackageinfo` tool. 

PkginfoMerger won't work for two reasons -- not all .munki recipes call it, and variable substitution is only handled for those keys that are in the Arguments array, or at the beginning of the run (and at the beginning env['url'] is undefined). 

You could make the variable substitution pathway work, but it would mean re-writing the .munki recipe to include pkginfo as an argument and not just doing a simple override. That's just not scalable across dozens of recipes.

paulsuh:
Correction - 'url' may not be defined at the start. It won't be if the recipe uses SparkleProvider or BrewCaskProvider or something like that. 

I also need to check to see if curl has a way to output the final download url in a clean way if it's following redirects.

nmcspadden:
@paulsuh are you planning to continue testing this?

nmcspadden:
Feel free to re-open if you want to pursue this.

PR #760, opened 2021-07-27
Fix issues related to Windows cURL Compatibility

jgstew:
URLGetter.py
- Fixed: cURL shipping on Windows 10 does not support `--compressed`
  - if curl binary path contains `Windows\\system32` then omit `--compressed`
  - adding a different curl to PATH would also fix this issue on windows, but this fix will not interfere with that case
  - Would be nice if I could somehow test the curl binary to see if it supports `--compressed` rather than assuming it doesn't in a very path specific way. This is a bigger issue if Windows ships a curl binary that does support `--compressed` in the future.
  - this fix makes getting up and running with AutoPkg on Windows easier, but might have negative implications for any recipe that requires `--compressed`, though I don't know if any actually would.

URLDownloaderPython.py
- black formatter changes due to removing pylint messages
- Fixed: was missing a file close handler which only caused issues on Windows
- changed the way headers are fetched to allow for future custom `User-Agent` headers which are required due to some software companies blocking cURL specifically.

### DIFF
--- a/Code/autopkglib/URLDownloaderPython.py
+++ b/Code/autopkglib/URLDownloaderPython.py
@@ -106,6 +106,10 @@ class URLDownloaderPython(URLDownloader):
             "required": False,
             "description": ("User Agent Header String to use for download"),
         },
+        "download_version": {
+            "required": False,
+            "description": ("version of download"),
+        },
     }
     output_variables = {
         "pathname": {"description": "Path to the downloaded file."},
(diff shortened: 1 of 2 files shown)


jgstew:
Also, I'm having issues related to this: https://github.com/autopkg/autopkg/pull/676

On Ubuntu / Windows.

jgstew:
I just realized there is a bunch of Windows specific enhancements in a different branch that is sitting out there: https://github.com/autopkg/autopkg/tree/dev_moved_processors_split_init_prefs_fixes 

Not sure what the plan is for reintegration into dev / master. Some of the changes seems more significant that others, and some seem to overlap with issues I have been dealing with and addressing myself.

PR #764, opened 2021-08-14
Resolve GitHubReleasesInfoProvider character decode error for Audacity

jgstew:
To resolve character decoding errors: https://stackoverflow.com/a/50709581/861745

This will remove and ignore the characters causing the issue from the results. Using `, encoding="utf-8"` also works, but not certain if that should be hard coded default or not?

This resolves the following issue:

```
The following recipes failed:
    com.github.jgstew.download.Audacity-Win
        Error in com.github.jgstew.download.Audacity-Win: Processor: GitHubReleasesInfoProvider: Error: 'charmap' codec can't decode byte 0x9d in position 85160: character maps to <undefined>
```

### DIFF
--- a/Code/autopkglib/github/__init__.py
+++ b/Code/autopkglib/github/__init__.py
@@ -272,7 +272,7 @@ To save the token, paste it to the following prompt."""
             self.http_result_code = int(header["http_result_code"])
 
         try:
-            with open(temp_content) as f:
+            with open(temp_content, errors="ignore") as f:
                 resp_data = json.load(f)
         except json.JSONDecodeError:
             resp_data = None


jgstew:
Recipe example:

```
Process:
    Arguments:
      github_repo: audacity/audacity
      asset_regex: .*64bit\.exe
      include_prereleases: False
```

homebysix (approved):

nmcspadden:
This feels like a much bigger change than it seems - ignoring errors on JSON load could have other profound implications for what happens if you load up a malformed file. I'd need to see significantly more testing of the various edge cases, and/or unit tests, before accepting this.

jgstew:
The other alternative is to set it as UTF8 as the default encoding, or set it to UTF8 and ignore errors.

My understanding is "ignore errors" will cause the characters that do not decode properly to be dropped from the output.

This seemed like the safest option with least impact since it would effectively remove non ascii characters from the output if an error is encountered, but do nothing in any existing case. 

I was less confident in making it always treat JSON files as UTF8 as a good default case.

"ignoring errors on JSON load could have other profound implications for what happens if you load up a malformed file"

Your concern is autopkg being pointed at an intentionally malicious file? That is interesting. I wonder if there is a way to ignore encoding errors WITHOUT ignoring JSON errors? Or maybe that is not sufficient?

Do you have example unit tests in mind? I was actually thinking of how to write some, potentially using a local web server to serve up examples for testing.

jgstew:
So interesting. Just checked. My recipe that is broken on Windows works fine on MacOS.

I suspect that on MacOS JSON decoding defaults to UTF8 always, while on Windows it defaults to ASCII Code Page 1252 - which is causing the issue.

I think providing an input variable to allow specifying a decoding method and making UTF8 the default on Windows makes the most sense.

nmcspadden:
That seems like a likely conclusion, because Windows just makes everything hard for no reason. I'm actually more inclined to make UTF8 the default because we can add that as an environmental assumption and document it - i.e. all JSON must be treated as UTF8 to be parsed. 

However, a variable that allows you to specify encoding and defaulting to UTF8 seems like the perfect compromise here that will allow us more flexibility later should we encounter some edge cases where we need to change it.

nmcspadden:
Just to come back to this: I wouldn't merge this PR as-is. I'm not comfortable with a blanket ignore_errors for JSON parsing. I'm okay with the suggestion for making the JSON parsing use UTF8 by default. If you want to hit up the stretch goal of making it a variable that can be tweaked, that's cool, but I wouldn't require it (I have yet to see anyone provide an example where this would cause a breakage).

jgstew:
So I think I found the source of the issue. On windows, you need to open the file as binary instead of the default of read as text. On non-windows it just works. I think it might be primarily a line endings issue, or it could just be that reading the file as binary forces python to use unicode as the encoding to read it as?

Basically doing this also fixes the issue on windows:

`with open(temp_content, "rb") as f:`

but I don't know if that could have a negative consequence for non-windows? I don't think so?

The main issue is that this will only work in Python 3.6+ as noted here: https://docs.python.org/3/library/json.html#json.load

"Changed in version 3.6: fp can now be a binary file. The input encoding should be UTF-8, UTF-16 or UTF-32."

jgstew:
Closing this pull request in favor of this approach: https://github.com/autopkg/autopkg/pull/783

That said, the discussion and background above in this PR is helpful.

PR #778, opened 2021-12-01
Fix for xattr usage on Linux

gregneagle:
Linux requires xattr names to start with "user." This fix allows autopkg on Linux to store extended attributes on downloaded files on some Linux-supported filesystems. ext4, for example, is known to work. Conversely, attempting to store extended attributes on files on NFS mounts is known to still fail.


nmcspadden:
Can you move this to the dev branch instead of master? Otherwise LGTM.

PR #783, opened 2022-01-05
Fix GitHubInfoReleasesProvider on Windows in case of non-ascii characters

jgstew:
This should resolve https://github.com/autopkg/autopkg/pull/764 (see discussion there)

This will allow GitHubInfoReleasesProvider to work on Windows for releases which contain unusual characters. It will require Python3.6+ on Windows, but will not induce that requirement for non-Windows. Reference: https://docs.python.org/3/library/json.html#json.load

I think this is a better approach since the encoding could be any of "UTF-8, UTF-16 or UTF-32" and this approach should work, rather than specifically forcing "UTF-8" to be the one used for decoding. From the footnote: "Changed in version 3.6: fp can now be a binary file. The input encoding should be UTF-8, UTF-16 or UTF-32."

This approach also has the advantage of not affecting the current typical behavior at all.

### DIFF
--- a/Code/autopkglib/github/__init__.py
+++ b/Code/autopkglib/github/__init__.py
@@ -274,12 +274,11 @@ To save the token, paste it to the following prompt."""
         try:
             with open(temp_content) as f:
                 resp_data = json.load(f)
+        except UnicodeDecodeError:
+            with open(temp_content, "rb") as f:
+                resp_data = json.load(f)
         except json.JSONDecodeError:
-            try:
-                with open(temp_content, "rb") as f:
-                    resp_data = json.load(f)
-            except json.JSONDecodeError:
-                resp_data = None
+            resp_data = None
 
         return (resp_data, self.http_result_code)
 
--- a/Code/autopkglib/github/__init__.py
+++ b/Code/autopkglib/github/__init__.py
@@ -271,6 +271,7 @@ To save the token, paste it to the following prompt."""
         if header["http_result_code"] != "000":
             self.http_result_code = int(header["http_result_code"])
 
+        resp_data = None
         try:
             with open(temp_content) as f:
                 resp_data = json.load(f)
@@ -278,7 +279,7 @@ To save the token, paste it to the following prompt."""
             with open(temp_content, "rb") as f:
                 resp_data = json.load(f)
         except json.JSONDecodeError:
-            resp_data = None
+            pass
 
         return (resp_data, self.http_result_code)
 
--- a/Code/autopkglib/github/__init__.py
+++ b/Code/autopkglib/github/__init__.py
@@ -275,7 +275,11 @@ To save the token, paste it to the following prompt."""
             with open(temp_content) as f:
                 resp_data = json.load(f)
         except json.JSONDecodeError:
-            resp_data = None
+            try:
+                with open(temp_content, "rb") as f:
+                    resp_data = json.load(f)
+            except json.JSONDecodeError:
+                resp_data = None
 
         return (resp_data, self.http_result_code)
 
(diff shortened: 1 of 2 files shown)


jgstew:
hmm, this might not be catching the correct exception? odd.

jgstew:
Now the correct exception is being handled.

jgstew:
without the code change in this PR, the following download recipe only works on non-windows, with this PR, it works on windows as well: https://github.com/jgstew/jgstew-recipes/blob/main/Audacity/Audacity-Win.download.recipe.yaml

Code/autopkglib/github/__init__.py
```diff
@@ -274,6 +274,9 @@ def call_api(
         try:
             with open(temp_content) as f:
                 resp_data = json.load(f)
+        except UnicodeDecodeError:
+            with open(temp_content, "rb") as f:
+                resp_data = json.load(f)
         except json.JSONDecodeError:
             resp_data = None
```
jgstew, line 281: maybe `resp_data` should be set to `None` before this try/except block on Line 273? Then this last except could just be a `pass`?

nmcspadden: Good catch, that's ugly code. That definitely doesn't work for all the cases.

nmcspadden:
Does this only affect Windows, or does this also affect macOS?

I'm wondering what the risks are of just using 'rb' unilaterally.

nmcspadden:
Actually, there's a much easier way to do this.

```
        resp_data = None
        read_mode = "rb" if is_windows else "r"
        try:
            with open(temp_content, read_mode) as f:
                resp_data = json.load(f)
        except json.JSONDecodeError:
            pass
```

I just tested this on TextMate2.download, which uses the GitHub processor. Can you test this and see if it works for your Windows recipe?

jgstew:
The reason `rb` is needed on windows is due to linebreaks being different when reading in text mode, but I think using `rb` might actually fix the issue in other cases that are similar. It seems like `rb` works on all OSes because `json.load` knows how to handle it correctly if binary. It just so happens that `r` in text mode does work on Unix style systems... probably in all cases, but who knows?

jgstew:
Something to note, I only get this error on windows for certain github repos, only those that use unicode in their release notes or titles.

jgstew:
Here is a failing case of the unmodified code with Audacity as the example:

```
GitHubReleasesInfoProvider
{'Input': {'asset_regex': '.*64bit\\.exe$',
           'github_repo': 'audacity/audacity',
           'include_prereleases': False}}
GitHubReleasesInfoProvider: No value supplied for CURL_PATH, setting default value of: /usr/bin/curl
GitHubReleasesInfoProvider: No value supplied for GITHUB_URL, setting default value of: https://api.github.com
GitHubReleasesInfoProvider: No value supplied for GITHUB_TOKEN_PATH, setting default value of: ~/.autopkg_gh_token
Receipt written to C:\Users\runneradmin/Library/AutoPkg/Cache\com.github.jgstew.test.AutoPkgCore\receipts\AutopkgCore.test.recipe-receipt-20220805-204637.plist

The following recipes failed:
    Test-Recipes/AutopkgCore.test.recipe.yaml
        Error in com.github.jgstew.test.AutoPkgCore: Processor: GitHubReleasesInfoProvider: Error: 'charmap' codec can't decode byte 0x9d in position 206596: character maps to <undefined>
```

It only fails on windows.

I should write a github action to test this.

jgstew:
Since GitHubReleaseInfoProvider uses network connectivity, the github action that tests it is only run if it itself is changed at the moment, and only on pull request, not push.

This run shows the succeeding test on this new code. 

This shows the failing test on the old code: https://github.com/jgstew/jgstew-recipes/actions/runs/2806149049

PR #784, opened 2022-01-05
Update xattr.py to not throw hard errors on some Linux systems.

jgstew:
xattr does not work in some Linux file systems, or in Linux running under WSL, or in some docker containers in general. This will make the failures of xattr no longer hard errors in these cases.

Use of xattr for storing etag data is not ideal in the first place and is already solved in URLDownloaderPython by using a json metadata file instead, but this will help with compatibility in general for older methods. This will have the drawback of not detecting that the already downloaded file is the same correctly due to lack of metadata for any recipe currently relying on xattr. That said, redownloading the file seems better than throwing hard errors.

I believe the behavior should be similar to how these older methods work on Windows today since on Windows xattr is no-op within autopkg.

### DIFF
--- a/Code/autopkglib/xattr.py
+++ b/Code/autopkglib/xattr.py
@@ -86,8 +86,7 @@ def getxattr(path: str, attr: str, symlink: bool = False) -> Optional[str]:
     try:
         return _xattr.getxattr(path, attr, symlink)
     except OSError as e:
-        print("WARNING: xattr.getxattr threw OSError.")
-        print(e)
+        print(f"WARNING: xattr.getxattr threw OSError. {e}")
         return None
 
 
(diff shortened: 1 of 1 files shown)


jgstew:
This should prevent hard errors in the case of "attempting to store extended attributes on files on NFS mounts is known to still fail": https://github.com/autopkg/autopkg/pull/778

Code/autopkglib/xattr.py
```diff
@@ -83,18 +83,38 @@ def setxattr(
 
 
 def getxattr(path: str, attr: str, symlink: bool = False) -> Optional[str]:
-    return _xattr.getxattr(path, attr, symlink)
+    try:
+        return _xattr.getxattr(path, attr, symlink)
+    except OSError as e:
+        print("WARNING: xattr.getxattr threw OSError.")
+        print(e)
```
nmcspadden, line 90: You could combine this:
```
print(f"WARNING: xattr.getxattr threw OSError: {e}")
```

nmcspadden:
Wait, didn't this get merged?

jgstew:
combined xatter error messaging as requested @nmcspadden

PR #795, opened 2022-04-09
remove incompatibility notice non-macs

jgstew:
remove notice about autopkg incompatibility on non-macs

### DIFF
--- a/Code/autopkg
+++ b/Code/autopkg
@@ -60,14 +60,6 @@ from autopkglib import (
 )
 from autopkglib.github import GitHubSession, print_gh_search_results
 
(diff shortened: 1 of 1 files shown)


jgstew:
It is fitting that I was the one that added this code originally when I first started experimenting with AutoPkg on non-Macs 7 years ago: https://github.com/autopkg/autopkg/commit/d097e9f926f86b9537255693e52fa9ae5f99db67

jgstew:
It probably should be the case that the NSFoundation import failures only print messages about them on MacOS, while on non-MacOS they should NOT print failure messages because they are never expected to work.

PR #799, opened 2022-04-14
new core processor: FolderCreator

discentem:
I was personally surprised to find that there is no `FolderCreator` processor, though there is `FileCreator`. Many recipes seem to use `PkgRootCreator` in face of `FolderCreator` not existing, like in this [Zoom recipe](https://github.com/autopkg/MLBZ521-recipes/blob/b8d38cb0332a759469e0a4b45eb388b729c4a724/Zoom/Zoom-ForIT.pkg.recipe#L113). 

Hence this PR :) It's secretly a stub for `PkgRootCreator` but I think it will be more intuitive for new recipe authors who need to create some folders.

### DIFF
--- a/Code/autopkglib/FolderCreator.py
+++ b/Code/autopkglib/FolderCreator.py
@@ -29,13 +29,13 @@ class FolderCreator(PkgRootCreator):
     input_variables = {
         "root": {
             "required": True,
-            "description": "Path to where the package root will be created.",
+            "description": "Path to where the root folder will be created.",
         },
         "subdirs": {
             "required": False,
             "description": (
                 "A dictionary of directories to be created "
-                "inside the pkgroot, with their modes in octal form."
+                "inside the root, with their modes in octal form."
             ),
         },
     }
--- a/Code/autopkglib/FolderCreator.py
+++ b/Code/autopkglib/FolderCreator.py
@@ -1,0 +1,55 @@
+#!/usr/local/autopkg/python
+#
+# Copyright 2010 Per Olofsson
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""See docstring for FolderCreator class"""
+
+import os.path
+import shutil
+
+from autopkglib.PkgRootCreator import PkgRootCreator
+
+__all__ = ["FolderCreator"]
+
+class FolderCreator(PkgRootCreator):
+    """Create a directory and relative subdirectories and set their permissions."""
+
+    description = __doc__
+    input_variables = {
+        "root": {
+            "required": True,
+            "description": "Path to where the package root will be created.",
+        },
+        "subdirs": {
+            "required": False,
+            "description": (
+                "A dictionary of directories to be created "
+                "inside the pkgroot, with their modes in octal form."
+            ),
+        },
+    }
+    output_variables = {}
+
+    def main(self):
+        root = self.env['root']
+        if 'subdirs' not in self.env:
+            subdirs = {}
+        else:
+            subdirs = self.env['subdirs']
+        # calls create on PkgRootCreator class
+        self.Create(root, subdirs)
+
+if __name__ == "__main__":
+    PROCESSOR = FolderCreator()
+    PROCESSOR.execute_shell()
--- a/Code/autopkglib/PkgRootCreator.py
+++ b/Code/autopkglib/PkgRootCreator.py
@@ -47,28 +47,28 @@ class PkgRootCreator(Processor):
     }
     output_variables = {}
 
-    def main(self):
+    def Create(self, pkgroot, pkgdirs):
         # Delete pkgroot if it exists.
         try:
-            if os.path.islink(self.env["pkgroot"]) or os.path.isfile(
-                self.env["pkgroot"]
+            if os.path.islink(pkgroot) or os.path.isfile(
+                pkgroot
             ):
-                os.unlink(self.env["pkgroot"])
-            elif os.path.isdir(self.env["pkgroot"]):
-                shutil.rmtree(self.env["pkgroot"])
+                os.unlink(pkgroot)
+            elif os.path.isdir(pkgroot):
+                shutil.rmtree(pkgroot)
         except OSError as err:
-            raise ProcessorError(f"Can't remove {self.env['pkgroot']}: {err.strerror}")
+            raise ProcessorError(f"Can't remove {pkgroot}: {err.strerror}")
 
         # Create pkgroot. autopkghelper sets it to root:admin 01775.
         try:
-            os.makedirs(self.env["pkgroot"])
-            self.output(f"Created {self.env['pkgroot']}")
+            os.makedirs(pkgroot)
+            self.output(f"Created {pkgroot}")
         except OSError as err:
-            raise ProcessorError(f"Can't create {self.env['pkgroot']}: {err.strerror}")
+            raise ProcessorError(f"Can't create {pkgroot}: {err.strerror}")
 
         # Create directories.
-        absroot = os.path.abspath(self.env["pkgroot"])
-        for directory, mode in sorted(self.env["pkgdirs"].items()):
+        absroot = os.path.abspath(pkgroot)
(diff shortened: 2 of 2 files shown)


nmcspadden:
What's the use case for this processor? Where would you use it vs. PkgRootCreator? What makes this a more appealing option for doing a subset of the functionality PkgRootCreator does?

discentem:
> What's the use case for this processor?

It took me multiple hours of searching through other people's recipes figure out that `PkgRootCreator` could be used to just create arbitrary folders. The use case of `FolderCreator` is to make it obvious to new recipe authors how they might create arbitrary folders. 

> Where would you use it vs. PkgRootCreator?

They are technically interchangeable, but there are many cases where it's not intuitive/obvious that `PkgRootCreator` solves the problem at hand. 
> What makes this a more appealing option...

It's more intuitive if someone needs to create an arbitrary folder and isn't trying to create a package. 

> ... for doing a subset of the functionality PkgRootCreator does?

Did I seriously flub the implementation? I really meant for this to just be a wrapper for `PkgRootCreator`. I don't think I broke any PkgRootCreator feature?

homebysix:
Setting aside for a moment whether a more discoverable FolderCreator processor is a good idea... one concern I would have is that PkgRootCreator uses the `pkgroot` output variable, which is conventionally used by downstream processors to perform other actions. I wouldn't want a hypothetical FolderCreator processor to change the `pkgroot` unexpectedly.

discentem:
> Setting aside for a moment whether a more discoverable FolderCreator processor is a good idea

Heh, I would appreciate your take though :) 

> one concern I would have is that PkgRootCreator uses the pkgroot output variable, which is conventionally used by downstream processors to perform other actions. I wouldn't want a hypothetical FolderCreator processor to change the pkgroot unexpectedly.

This is a really good call out. I had not considered this. I will go back to the code and sort this out as I agree: FolderCreator should output to separate output variables.

discentem:
@homebysix I think we should decide first if there is an appetite for this new processor and then discuss implementation. But, with regards to implementation: 

I actually have been unable to spot where `PkgRootCreator` outputs to `pkgroot`? 

Currently, as of [this commit ](https://github.com/autopkg/autopkg/blob/85cb634992f52d8c975e6350e0b9df453686d74b/Code/autopkglib/FolderCreator.py), `FolderCreator` uses separate input variables so it would not risk overwriting `self.env['pkroot']`. I feat I am missing something though.

discentem:
Here's the output of my usage of both `FolderCreator` and `PkgRootCreator` in a single recipe: 

**Output**: https://gist.github.com/discentem/1e62382f326d19383e5c7302937a3070
**Recipe**: https://github.com/discentem/discentem-recipes/blob/42affba94e616bdf33eb2c664239500540af3e1d/Bazel/bazelisk.munki.recipe.yaml
**Parent Recipes**: https://github.com/discentem/discentem-recipes/tree/42affba94e616bdf33eb2c664239500540af3e1d/Bazel

arubdesu:
It would be helpful, for gauging when to reach for this processor, to know what current recipes are 'overloading' the original intent of pkgrootcreator. @discentem I see your bazel use case (do you really need the postinstall script in the pkg recipe... when you could make it executable in the payload you just assembled?😜), what problems are the people getting by with the pre-existing processor achieving in their recipes? Are you making a dmg with a specific directory layout in it?

It seems for your (and the original) use case we could achieve the same goal by rebranding/renaming the processor pkgrootdirhandler, but I get caught up on semantics and don't have skin in the game to vote one way or the other. ('2 hard problems in computer science' joke and all that.) And we'd probably want/need code for stub'ing/transitioning users if the rebranding were how we decided to handle this concern just the same.

Although we tend to leave it up to some guesswork that the majority of processors only have code comments and docstrings on input/outputs as documentation, I'd just like to square what scenarios this would be employed in, even if they're relatively small in total number.

discentem:
Hey @arubdesu, thanks for the reply! Will try to cover all of your questions, please shout if I miss anything. 

> It would be helpful, for gauging when to reach for this processor, to know what current recipes are 'overloading' the original intent of pkgrootcreator.

I kind of disagree here. I don't intend to ask anyone to move existing recipes to `FolderCreator`, even if they are indeed 'overloading' `PkgRootCreator`. It isn't necessary as their recipes are not broken. 

However, if requested, I can try to do some research and see how many recipes beyond the [Zoom recipe](https://github.com/autopkg/MLBZ521-recipes/blob/b8d38cb0332a759469e0a4b45eb388b729c4a724/Zoom/Zoom-ForIT.pkg.recipe#L113) perform this sort of overloading. 

> I see your bazel use case (do you really need the postinstall script in the pkg recipe... when you could make it executable in the payload you just assembled?😜)... 

I swear that I tried to do this without a postinstall but, if I recall correctly, Copier or PkgRequest did not appear to transfer the permissions all the way through. I can give it another shot but that is not strictly related to FolderCreator IMO (in my opinion). 

> what problems are the people getting by with the pre-existing processor achieving in their recipes? 

Without doing more research, simply stated: in any scenario where people need to create an arbitrary folder they often appear to be using `PkgRootCreator`.

> It seems for your (and the original) use case we could achieve the same goal by rebranding/renaming the processor pkgrootdirhandler, but I get caught up on semantics and don't have skin in the game to vote one way or the other. ('2 hard problems in computer science' joke and all that.)

There is no particular reason I've stubbed `PkgRootCreator` instead of `PkgRootDirHandler` except that (a) I've never used `PkgRootDirHandler` and (b) when I was trying to solve getting a postinstall in the right place I looked to other recipes for inspiration: the Zoom recipe solved it with `PkgRootCreator`. 

> And we'd probably want/need code for stub'ing/transitioning users if the rebranding were how we decided to handle this concern just the same.

What did you have in mind? I don't currently see it as necessary to convert any existing recipes to use `FolderCreator`. I just see `FolderCreator` as a way to help people with net-new recipes. 

> I'd just like to square what scenarios this would be employed in, even if they're relatively small in total number.

This is interesting. If there is a rule or convention about only adding Core Processors that are proven to be broadly useful than perhaps I should do the research that I've been hesitant to do. It just seems obvious to me that if `FileCreator` exists, `FolderCreator` should also exist. But if this isn't good enough but itself and further research/proof is requested, I will definitely give that consideration if I want to move forward.

nmcspadden:
The reason I think these questions are important is that there's a sort of unwritten expectation that a Core processor would generally be something base level and applicable to everyone. If there's no existing need for this processor in existing recipes - because the existing functionality suffices - then I'm not super clear on the value this brings. It may seem obvious to have a simple FolderCreator processor conceptually, but less code is usually better, and if there's no additional value-add from FolderCreator in existing recipes compared to PkgRootCreator, then I really have to question the importance.

Part of the problem is that we don't have any written guidelines for what should or shouldn't be a Core processor. About the only guarantee is that AutoPkg maintainers are expected to support their core processors (compared to shared processors).

discentem:
> The reason I think these questions are important is that there's a sort of unwritten expectation that a Core processor would generally be something base level and applicable to everyone.

I agree and understand that these questions are important. Even though existing users of `PkgRootCreator` would not be required to migrate to `FolderCreator` I still think `FolderCreator` applies to everyone, but especially new recipe authors.

> If there's no existing need for this processor in existing recipes - because the existing functionality suffices - then I'm not super clear on the value this brings.

It offers, a better user experience to people who are new to recipe writing that need to create folders. I personally see it as a plus that it does not break any existing code. 

> if there's no additional value-add from FolderCreator in existing recipes compared to PkgRootCreator, then I really have to question the importance.

"Improved User Experience for new recipe writers" might not be enough reason to merge this thing and, if it's not, that's fine with me. I could just as well ship it as a shared processor. 

**Alternatively**, I could create an entry on the wiki for `FolderCreator` (a processor which does not exist) and write a little blurb about how `PkgRootCreator` can accomplish the task. This would require zero code change and would have improved my recipe writing experience. 

> Part of the problem is that we don't have any written guidelines for what should or shouldn't be a Core processor. About the only guarantee is that AutoPkg maintainers are expected to support their core processors (compared to shared processors).

I absolutely don't want to increase maintainer burden for anything that is not considered "worth it" and empathize with the fact that no guidelines exist. If you say "we aren't going to accept this" that's fine with me. I'd prefer that over a drawn out debate. Drawn-out debates can be superbly important to open-source projects, this change isn't terribly consequential.

nmcspadden:
Just to close the loop on this, I'm still not convinced this is needed at this time. While I'm certainly open to adding more core processors that could be useful, I would really need to see this fulfill a use case that isn't currently being met. So far, there just doesn't seem to be a burning need for this, especially since everyone else has already decided on a different course of action.

It doesn't quite seem like switching from PkgRootCreator to this actually results in significantly simpler (although that's a little hard to quantify) or shorter recipes, which really begs the question for how necessary it is.

That being said, I want to continue to encourage you (and others) to put forth PRs or Issues with these kinds of ideas to generate exactly this kind of discussion. Thanks for your contribution to the community discussion!

discentem:
❤️

PR #800, opened 2022-04-20
Implement xattr support on Windows

NickETH:
Implementing xattr support on Windows by gluing ETAG to File with ADS (alternate data stream).
Changing the config file loacation to %APPDATA%, instead of %LOCALAPPDATA%. This change is not strictly necessary, but would be good policy.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,5 @@
+## [2.7.7](https://github.com/autopkg/autopkg/compare/v2.7.6...HEAD) (Unreleased)
+
 # AutoPkg Change Log
 
 All notable changes to this project will be documented in this file. This project adheres to [Semantic Versioning](http://semver.org/).
--- a/Code/autopkglib/version.plist
+++ b/Code/autopkglib/version.plist
@@ -3,6 +3,6 @@
 <plist version="1.0">
 <dict>
 	<key>Version</key>
-	<string>2.7.6</string>
+	<string>2.7.7</string>
 </dict>
 </plist>


nmcspadden:
This seems reasonable, but can you please provide some output from AutoPkg on both macOS and Windows to show that this actually works as intended? I'd like to see what the relevant output is, where the data is stored on Windows, and how it tracks etags.

Code/autopkg
```diff
@@ -806,6 +807,9 @@ def repo_delete(argv):
             recipe_search_dirs.remove(repo_path)
         # now remove the repo files
         try:
+            if is_windows():
+                # we need to remove readonly, hidden and system bits
+                os.system("attrib.exe -r -h -s " + repo_path + "\\*.* /S /D")
```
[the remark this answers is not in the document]
NickETH: Just simplicity, i think. It is working with os.system, but could of course use subprocess too.

NickETH:
Ok, here is an output from a Windows machine:
https://gist.github.com/NickETH/d99ce98224e01189cc893995bce59762
Do not have a Mac around here, but could ask Graham next week to do it.
What is the reason, the unittests failed? Did not make any changes in the part, where it complains.

NickETH:
Hi Nick,
do you see a chance to correct this failing test in the next few days?
Would be cool, to get this done.
Thanks

jgstew:
The step of the test that is failing is the e2e test, running the following:

`python Code/autopkg info Firefox.munki --prefs tests/preferences.plist`

and it is not finding the `Firefox.munki` recipe. I am not sure why. I wonder if it has to do with `--prefs tests/preferences.plist`?

jgstew:
Also, this PR has changes in it that are NOT only "Implement xattr support on Windows" and seems to contain 3 different sets of changes.

Code/autopkglib/pyads/pyads.py
```diff
@@ -0,0 +1,138 @@
...
+to the warranties of merchantability, fitness for a particular purpose and noninfringement. In no event shall
+the authors or copyright holders X be liable for any claim, damages or other liability, whether in an action
+of contract, tort or otherwise, arising from, out of or in connection with the software or the use or other
+dealings in the Software.
+Except as contained in this notice, the name of the Robin David shall not be used in advertising or otherwise
+to promote the sale, use or other dealings in this Software without prior written authorization from the Robin David.
+"""
+import os
+import sys
+from ctypes import *
```
jgstew, line 19: This doesn't belong as a file in the project itself, but rather a pip dependency.

Code/autopkg
jgstew, line 812: This seems to be to fix a bug that is unrelated to fixing xattr. What is the issue this is solving? Is there a pythonic way to solve this using Pathlib or similar, rather than calling `attrib.exe`?

I use autopkg a ton on windows, and I have never hit the problem that this seems to be fixing. What autopkg command causes this problem?

Code/autopkglib/StopProcessingIf.py
```diff
@@ -47,14 +55,51 @@ class StopProcessingIf(Processor):
...
+                raise ProcessorError(f"Predicate error for '{predicate_string}': {err}")
+
+            result = predicate.evaluateWithObject_(self.env)
+            self.output(f"({predicate_string}) is {result}")
+            return result
+        elif is_windows():  # Added on Windows version
+            result = False
+            try:
+                # result = eval(predicate_string)
+                splitted_ops = predicate_string.split()
```
jgstew, line 72: I think this is assuming that predicates have a space between the left side, the right side, and the operator(s) but I don't think that is always the case. Is `download_changed!=True` a valid predicate? I don't actually have enough experience with them to know for sure.

Code/autopkglib/WindowsSignatureVerifier.py
```diff
@@ -0,0 +1,112 @@
...
+                "Subject"
+            ),
+        },
+    }
+    output_variables = {}
+
+    description = __doc__
+
+    def main(self):
+        if not is_windows():
```
jgstew, line 70: Can powershell core do `Get-AuthenticodeSignature` on non-windows? I'm not certain, but if so, then the decision not to run this might have more to do with finding a powershell binary to execute than it is to do with `is_windows()`.

I have never tried to use `Get-AuthenticodeSignature` in Powershell core on non-windows, but I should try it.

Code/autopkglib/WindowsSignatureVerifier.py
```diff
@@ -0,0 +1,112 @@
...
+            self.output("Not on Windows, not running Windows Signature " "Verifier")
+            return
+        if self.env.get("DISABLE_CODE_SIGNATURE_VERIFICATION"):
+            self.output("Code signature verification disabled for this recipe " "run.")
+            return
+        input_path = self.env["input_path"]
+        powershell = "C:\\windows\\System32\\WindowsPowerShell\\v1.0\\powershell.exe"
```
jgstew, line 77: This is a very minor point, not a blocker, but I like the idea of the powershell path being an input variable, but with `"C:\\windows\\System32\\WindowsPowerShell\\v1.0\\powershell.exe"` being the default path.

I like to use a Processor that takes an array of paths and looks for the first one it find that is an executable file and returns it, then passes that as input into the next processor like this one, which makes it easier to implement this using PowerShell core instead of Powershell, or non-windows if that works, which I haven't checked to see if it does.

Code/autopkglib/__init__.py
```diff
@@ -216,11 +216,11 @@ def _get_macos_prefs(self):
...
     def _get_file_prefs(self):
         r"""Lookup preferences for Windows in a standardized path, such as:
-        * `C:\\Users\username\AppData\Local\Autopkg\config.{plist,json}`
+        * `C:\Users\username\AppData\Roaming\Autopkg\config.{plist,json}`
         * `/home/username/.config/Autopkg/config.{plist,json}`
         Tries to find `config.plist`, then `config.json`."""
-
-        config_dir = appdirs.user_config_dir(APP_NAME, appauthor=False)
+        # Windows configuration files should go to the Appdata\Roaming dir.
+        config_dir = appdirs.user_config_dir(APP_NAME, appauthor=False, roaming=True)
```
jgstew, line 223: I wonder if it should be checking roaming first, but then check local as a fallback?

I think this change would break my autopkg prefs on my windows devices until I moved the prefs file from local to roaming.

Code/autopkglib/pyads/pyads.py
NickETH: Bad idea...
This module does a complete different thing (it is for PLC API)!
Unfortunateley, they are both named the same.
There is no module for pyads for alternate data streams, i am aware of.

Code/autopkglib/StopProcessingIf.py
jgstew: This isn't a true replacement for `StopProcessingIf`'s use of predicates, but it does work the same way cross platform: https://github.com/jgstew/jgstew-recipes/blob/main/SharedProcessors/StopProcessingIfDownloadUnchanged.py

Code/autopkg
NickETH: This changes came into this PR, after i issued it.
Was not intend to do so...

jgstew: ah, I see. That is a pain. Once you create the PR from a branch, if you make any further changes to that branch, then it gets included. 

It is best to create a custom branch only for the purposes of the PR, which is a bit annoying.

nmcspadden:
Honestly, I would prefer not to use an equivalent to xattr on Windows. Let's keep it simple and just store the etag information in a JSON file on disk, and read/write to it. Embedding it in the download just to mimic what it does on macOS using extended attributes seems unnecessarily difficult.

Code/autopkglib/pyads/pyads.py
jgstew: oh you are right, this is a different pip library. So where is this one you are using from?

jgstew:
This currently requires switching over to URLDownloaderPython on windows and non windows in order to achieve the JSON option, but an equivalent JSON file could be used as a fallback on URLDownloader as well, not just on windows, but on any file system that doesn't support xattr.

NickETH:
> Honestly, I would prefer not to use an equivalent to xattr on Windows. Let's keep it simple and just store the etag information in a JSON file on disk, and read/write to it. Embedding it in the download just to mimic what it does on macOS using extended attributes seems unnecessarily difficult.

I see your point.
That is the future, i think and should be implemented globally, IMHO.
But in the meantime, this solution works pretty reliable.
Using it in production on a daily basis.

nmcspadden:
I'd like to potentially revisit this concept for a future AutoPkg release. Given what we have in 3.0.0RC1 right now, what would need to be done still to accommodate this?

NickETH:
did not follow close on your work in the last few weeks.
Let me get the 3.0 code and test.
But i think, it would still blend in without changes.
Will come back, as soon as i have tested it.

NickETH:
Looks like, the PR would work as expected.
Should i do a new one on 3.0RC1?

PR #810, opened 2022-06-04
Add plist serializer

MLBZ521:
tl/dr:  This simple function should resolve an (edge case) issue where `plistlib.dump` raises an exception when the contents contain `None` type values.

Details:
YAML supports `null` values, as does Python `dict`s as `None`, however, Plists do not and [`plistlib.dump()`](https://docs.python.org/3/library/plistlib.html#plistlib.dump) raises a `TypeError` if a value is null.

I have recipes overrides where I simply override specific keys with an empty value, e.g.:
```
<key>OS_REQUIREMENTS</key>
<string></string>
```

This has worked without issue for years.  But recently, I've began creating new overrides and converting current overrides to YAML when/if I'm update them.  Following the same premise, the empty value could be accomplished by:
```
  OS_REQUIREMENTS: null
```
or
```
  OS_REQUIREMENTS: 
```
All the the libraries and existing logic within AutoPkg _actually_ supports both above options (so much so, that `autopkg update-trust-info <recipe_id>` leaves `null` values in place, as would likely be expected).....all the way up to when the contents of the `run_results` variable is passed into `plistlib.dump()` on an `autopkg run <recipe_id>` and then an error occurs:
```
{
    "Traceback (most recent call last):
  File \"/usr/local/bin/autopkg\", line 2758, in <module>
    sys.exit(main(sys.argv))
  File \"/usr/local/bin/autopkg\", line 2754, in main
    return subcommands[verb][\"function\"](argv)
  File \"/usr/local/bin/autopkg\", line 2276, in run_recipes
    plistlib.dump(run_results, f)
  File \"/Library/AutoPkg/Python3/Python.framework/Versions/Current/lib/python3.7/plistlib.py\", line 970, in dump
    writer.write(value)
  File \"/Library/AutoPkg/Python3/Python.framework/Versions/Current/lib/python3.7/plistlib.py\", line 402, in write
    self.write_value(value)
  File \"/Library/AutoPkg/Python3/Python.framework/Versions/Current/lib/python3.7/plistlib.py\", line 437, in write_value
    self.write_array(value)
  File \"/Library/AutoPkg/Python3/Python.framework/Versions/Current/lib/python3.7/plistlib.py\", line 482, in write_array
    self.write_value(value)
  File \"/Library/AutoPkg/Python3/Python.framework/Versions/Current/lib/python3.7/plistlib.py\", line 437, in write_value
    self.write_array(value)
  File \"/Library/AutoPkg/Python3/Python.framework/Versions/Current/lib/python3.7/plistlib.py\", line 482, in write_array
    self.write_value(value)
  File \"/Library/AutoPkg/Python3/Python.framework/Versions/Current/lib/python3.7/plistlib.py\", line 425, in write_value
    self.write_dict(value)
  File \"/Library/AutoPkg/Python3/Python.framework/Versions/Current/lib/python3.7/plistlib.py\", line 472, in write_dict
    self.write_value(value)
  File \"/Library/AutoPkg/Python3/Python.framework/Versions/Current/lib/python3.7/plistlib.py\", line 425, in write_value
    self.write_dict(value)
  File \"/Library/AutoPkg/Python3/Python.framework/Versions/Current/lib/python3.7/plistlib.py\", line 472, in write_dict
    self.write_value(value)
  File \"/Library/AutoPkg/Python3/Python.framework/Versions/Current/lib/python3.7/plistlib.py\", line 440, in write_value
    raise TypeError(\"unsupported type": {
        "%s\" % type(value))
TypeError": {
            "unsupported type": "<class 'NoneType'>"
        }
    }
}
```
The recipe run _actually seems_ to complete successfully, but this error is produced after the fact (at least, my wrapper around `autopkg` produces a success message and then an error message, strangely enough).

I've only see this using the `run` subcommand, but I went ahead and added the function to all uses of `plistlib.dump`.

So far, I've found [one other report](https://macadmins.slack.com/archives/C056155B4/p1632752937261400) of this, though cannot confirm it is due to this same issue.

### DIFF
--- a/Code/autopkg
+++ b/Code/autopkg
@@ -587,7 +587,7 @@ def write_plist_exit_on_fail(plist_dict, path):
     if the write fails."""
     try:
         with open(path, "wb") as f:
-            plistlib.dump(plist_dict, f)
+            plistlib.dump(plist_serializer(plist_dict), f)
     except (TypeError, OverflowError):
         log_err(f"Failed to save plist to {path}.")
         sys.exit(-1)
@@ -1741,7 +1741,7 @@ def update_trust_info(argv):
                     yaml.dump(recipe, f, encoding="utf-8")
             else:
                 with open(recipe_path, "wb") as f:
-                    plistlib.dump(recipe, f)
+                    plistlib.dump(plist_serializer(recipe), f)
             log(f"Wrote updated {recipe_path}")
         else:
             log_err(
@@ -1971,7 +1971,7 @@ def make_override(argv):
             yaml.dump(override_dict, f, encoding="utf-8")
     else:
         with open(override_file, "wb") as f:
-            plistlib.dump(override_dict, f)
+            plistlib.dump(plist_serializer(override_dict), f)
     log(f"Override file saved to {override_file}")
     return 0
 
@@ -2178,7 +2178,7 @@ def run_recipes(argv):
     run_results = []
     try:
         with open(current_run_results_plist, "wb") as f:
-            plistlib.dump(run_results, f)
+            plistlib.dump(plist_serializer(run_results), f)
     except OSError as err:
         log_err(f"Can't write results to {current_run_results_plist}: {err.strerror}")
 
@@ -2291,7 +2291,7 @@ def run_recipes(argv):
         run_results.append(autopackager.results)
         try:
             with open(current_run_results_plist, "wb") as f:
-                plistlib.dump(run_results, f)
+                plistlib.dump(plist_serializer(run_results), f)
         except OSError as err:
             log_err(
                 f"Can't write results to {current_run_results_plist}: {err.strerror}"
@@ -2354,7 +2354,7 @@ def run_recipes(argv):
             receipt_path = os.path.join(receipt_dir, receipt_name)
             try:
                 with open(receipt_path, "wb") as f:
-                    plistlib.dump(autopackager.results, f)
+                    plistlib.dump(plist_serializer(autopackager.results), f)
                 if options.verbose:
                     log(f"Receipt written to {receipt_path}")
             except OSError as err:
@@ -2683,10 +2683,29 @@ def new_recipe(argv):
                 yaml.dump(recipe, f, encoding="utf-8")
         else:
             with open(filename, "wb") as f:
-                plistlib.dump(recipe, f)
+                plistlib.dump(plist_serializer(recipe), f)
         log(f"Saved new recipe to {filename}")
     except Exception as err:
         log_err(f"Failed to write recipe: {err}")
+
+
+def plist_serializer(obj):
+    """Serialize an object to ensure it can be dumped in plist format.
+
+    Args:
+        obj (dict, list): Object is assumed to be either a dict or list
+            that will be parsed.
+
+    Returns:
+        (any): The received object will be returned, modified if required.
+    """
+    if isinstance(obj, dict):
+        for k, v in obj.items():
+            obj[k] = "" if v is None else plist_serializer(v)
+    elif isinstance(obj, list):
+        for item in range(len(obj)):
+            plist_serializer(obj[item])
+    return obj
 
 
 def main(argv):


MLBZ521:
```
$ git commit -m "Add plist serializer"
[WARNING] Unstaged files detected.
[INFO] Stashing unstaged files to /Users/mlbz521/.cache/pre-commit/patch1654333778.
black....................................................................Passed
isort....................................................................Passed
flake8...................................................................Passed
[INFO] Restored changes from /Users/mlbz521/.cache/pre-commit/patch1654333778.
```

MLBZ521:
This probably isn't required but...validating the TypeError example:
```
$ /usr/local/autopkg/python 
Python 3.7.5 (v3.7.5:5c02a39a0b, Oct 14 2019, 18:49:57) 
[Clang 6.0 (clang-600.0.57)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> 
>>> mydict = { "one":  1, "two":  None }
>>> mydict
{'one': 1, 'two': None}
>>> type(mydict)
<class 'dict'>
>>> import plistlib
>>> with open("/Users/Shared/mydict-test", "wb") as fp:
...     plistlib.dump(mydict, fp)
... 
  File "<stdin>", line 2, in <module>
  File "/Library/AutoPkg/Python3/Python.framework/Versions/Current/lib/python3.7/plistlib.py", line 970, in dump
    writer.write(value)
  File "/Library/AutoPkg/Python3/Python.framework/Versions/Current/lib/python3.7/plistlib.py", line 402, in write
    self.write_value(value)
  File "/Library/AutoPkg/Python3/Python.framework/Versions/Current/lib/python3.7/plistlib.py", line 425, in write_value
    self.write_dict(value)
  File "/Library/AutoPkg/Python3/Python.framework/Versions/Current/lib/python3.7/plistlib.py", line 472, in write_dict
    self.write_value(value)
  File "/Library/AutoPkg/Python3/Python.framework/Versions/Current/lib/python3.7/plistlib.py", line 440, in write_value
    raise TypeError("unsupported type: %s" % type(value))
TypeError: unsupported type: <class 'NoneType'>
```

Code/autopkg
```diff
@@ -2683,12 +2683,31 @@ def new_recipe(argv):
...
+    Args:
+        obj (dict, list): Object is assumed to be either a dict or list
+            that will be parsed.
+
+    Returns:
+        (any): The received object will be returned, modified if required.
+    """
+    if isinstance(obj, dict):
+        for k, v in obj.items():
+            obj[k] = "" if v is None else plist_serializer(v)
```
nmcspadden, line 2704: This feels slightly dangerous and could cause type safety violations down the line, but I don't have a better answer to this now.

Code/autopkg
```diff
@@ -2683,12 +2683,31 @@ def new_recipe(argv):
...
         else:
             with open(filename, "wb") as f:
-                plistlib.dump(recipe, f)
+                plistlib.dump(plist_serializer(recipe), f)
         log(f"Saved new recipe to {filename}")
     except Exception as err:
         log_err(f"Failed to write recipe: {err}")
 
 
+def plist_serializer(obj):
```
nmcspadden, line 2692: Can you move this into autopkglib instead and import it into autopkg from autopkglib?

I would like to eventually move as much code as possible out of the autopkg file and put more of it into the library.

nmcspadden (changes requested):
Thanks for doing this, this is a nice safety net to put in. My only request is to move the function to autopkglib and import it into autopkg.

Code/autopkg
MLBZ521: Are you specifically referring to `isinstance(obj, dict)`
or `"" if v is None else plist_serializer(v)`?

I do know the former is a debate whether or not it should be used, but in this case, I as well, cannot think of another option at the moment.  I know the original object `run_results` being sent to the function, is a list of objects, but i was attempting to not assume what it would always be (when used by other functions; i.e. attempting to keep it agnostic).

For the latter, I've also tried to think of what could be fouled up with that condition, but haven't been able to think of anything yet.

Code/autopkg
MLBZ521: I actually had this thought for fleeting moment when I was debating where to insert the function...  This makes sense.

Code/autopkg
nmcspadden: I was more concerned about the fact that any None value is inherently casted as an empty string instead. This is probably safe enough in our usage here, but as we add more and more type safety checks to AutoPkg's python code, this may need to be refactored a bit.

However, that is very much a "after Tomorrow" problem.

MLBZ521: Ah, I see.  I was thinking about that too, but I couldn't think of a way to otherwise infer what the value _should be_ to properly cast it into the corresponding plist/xml format.

nmcspadden:
Can you move this to autopkg:dev instead of master? Then I'll merge.

Sorry for the delay!

jgstew:
This is a nice fix. I have run into this issue so many times.

MLBZ521:
@nmcspadden Sorry for the delay, this has been done.

PR #816, opened 2022-08-02
Fix URLGetter.py when default windows curl is used.

jgstew:
Fix URLGetter for cases where default windows curl is used.


jgstew:
I'm not sure if the automated checks cover this case. I need to double check and test this.

PR #863, opened 2023-06-27
Allow .tar.xz files to be processed as .tar

homebysix:
When this change was initially [proposed](https://github.com/autopkg/autopkg/pull/164) in 2015, it was problematic because xz files weren't handled the same in all Python versions that shipped with macOS. Now that autopkg ships its own Python, we're in control of that.

Keeping in WIP status until I can gather previously nonworking use cases and post test results.


jgstew:
FYI, automated tests are failing due to: https://github.com/autopkg/autopkg/pull/897

homebysix:
Cherry-picked to 2.7.x-dev branch.

PR #897, opened 2023-09-26
fix github action unittests tests.yaml

jgstew:
fix autopkg tests by having it generate recipe map.


jgstew:
I'm not sure why the windows tests are still failing, it might be due to problems setting the github token and it not working correctly?

It is specifically SignToolVerifier that is failing on windows, but the rest of the tests are working now.

jgstew:
Note: `${{ secrets.GITHUB_TOKEN }}` is how github actions provide the actual token inline. That shouldn't change, but where it is written to should based upon the new default path for the github token.

jgstew:
The github token doesn't seem to be read correctly in the unit tests, and the default value doesn't seem to get written into the preferences like I expect:

AutoPkg_GH_Token

Issue #893, opened 2023-09-25
recipe map creation fails on new autopkg setup


jgstew:
**THIS IS ONLY INTENDED FOR AUTOPKG BETAS.**

**Describe the problem**
Clone autopkg repo, switch to dev branch, install dependancies, then try to run anything or do anything that would create autopkg recipe map file.

Like: `jgstew@jgstews-iMac autopkg % python3 ./Code/autopkg info`

**Preferences contents**
*BE SURE TO SANITIZE ANY SENSITIVE DATA SUCH AS PASSWORDS OR ADDRESSES.*

Not applicable, none created yet.

**AutoPkg output**
*BE SURE TO SANITIZE ANY SENSITIVE DATA SUCH AS PASSWORDS OR ADDRESSES.*

```
jgstew@jgstews-iMac autopkg % python3 ./Code/autopkg info
/Users/jgstew/Library/Python/3.9/lib/python/site-packages/urllib3/__init__.py:34: NotOpenSSLWarning: urllib3 v2.0 only supports OpenSSL 1.1.1+, currently the 'ssl' module is compiled with 'LibreSSL 2.8.3'. See: https://github.com/urllib3/urllib3/issues/3020
  warnings.warn(
Cannot read the recipe map file!
Cannot find or read the recipe map! Creating it now...
Traceback (most recent call last):
  File "/Users/jgstew/Projects/autopkg/./Code/autopkg", line 2725, in <module>
    sys.exit(main(sys.argv))
  File "/Users/jgstew/Projects/autopkg/./Code/autopkg", line 2721, in main
    return subcommands[verb]["function"](argv)
  File "/Users/jgstew/Projects/autopkg/./Code/autopkg", line 893, in get_info
    read_recipe_map(rebuild=True)
  File "/Users/jgstew/Projects/autopkg/Code/autopkglib/__init__.py", line 428, in read_recipe_map
    calculate_recipe_map()
  File "/Users/jgstew/Projects/autopkg/Code/autopkglib/__init__.py", line 344, in calculate_recipe_map
    write_recipe_map_to_disk()
  File "/Users/jgstew/Projects/autopkg/Code/autopkglib/__init__.py", line 384, in write_recipe_map_to_disk
    with open(DEFAULT_RECIPE_MAP, "w") as f:
FileNotFoundError: [Errno 2] No such file or directory: '/Users/jgstew/Library/AutoPkg/recipe_map.json'
```

**Expected behavior**

[...]

PR #901, opened 2023-09-28
Supporting reading DEFAULT_RECIPE_MAP from preferences

nstrauss:
In the current dev branch `DEFAULT_RECIPE_MAP` for recipe_map.json path only supports the default. Since it's somewhat common in CI workflows to not use `~/Library/AutoPkg` for recipe repo or override directories, also allow the recipe map location to be configurable.

Could potentially mitigate https://github.com/autopkg/autopkg/issues/893 as well when the preference is set before autopkg is run.

### DIFF
--- a/Code/autopkglib/__init__.py
+++ b/Code/autopkglib/__init__.py
@@ -283,6 +283,11 @@ def find_identifier_from_name(name: str) -> Optional[str]:
     return None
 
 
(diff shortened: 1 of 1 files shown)


homebysix:
Should be superseded by PR #1027, which is now in the dev branch and will be the basis of a future 3.0 release. Thanks for submitting!

PR #904, opened 2023-10-05
Use expand-full switch with pkgutil

marczak:
The (undocumented) expand-full switch will expand all components and better drill down into distribution-style packages.

### DIFF
--- a/Code/autopkglib/FlatPkgUnpacker.py
+++ b/Code/autopkglib/FlatPkgUnpacker.py
@@ -138,7 +138,7 @@ class FlatPkgUnpacker(DmgMounter):
         try:
             pkgutilcmd = [
                 "/usr/sbin/pkgutil",
-                "--expand",
+                "--expand-full",
                 self.source_path,
                 self.env["destination_path"],
             ]


marczak:
Addendum: I don't recall where this switch appeared, so, this could break the processor on much older OS Versions (10.11-era and earlier).

nmcspadden:
1. Please swap this to the dev branch.
2. What is this actually doing? What's different about the expanded output compared to before? Showing the results of an actual recipe run could help. I don't really have any context on what this is actually solving or accomplishing.

marczak:
Ah, sure. --expand-full will also expand a package's payload. Here's an example from distribution-style package:

with --expand:
`% ls -la                                                                                   
total 43096
drwx------  6 marg  staff       192 Oct  6 17:25 .
drwxr-xr-x  5 marg  staff       160 Oct  6 17:25 ..
-rw-r--r--  1 marg  staff     38866 Sep 28 19:08 Bom
-rw-r--r--  1 marg  staff       933 Oct  6 17:25 PackageInfo
-rw-r--r--  1 marg  staff  21641179 Sep 28 19:08 Payload
drwxr-xr-x  7 marg  staff       224 Oct  6 17:25 Scripts`

same package with --expand-full:
`% ls -la                                                                                   
total 88
drwxr-xr-x  6 marg  staff    192 Oct  6 17:26 .
drwxr-xr-x  5 marg  staff    160 Oct  6 17:26 ..
-rw-r--r--  1 marg  staff  38866 Sep 28 19:08 Bom
-rw-r--r--  1 marg  staff    933 Oct  6 17:26 PackageInfo
drwxr-xr-x  4 marg  staff    128 Oct  6 17:26 Payload
drwxr-xr-x  7 marg  staff    224 Oct  6 17:26 Scripts`

Notice that in the second example, Payload is now a directory, requiring no further expansion. In the case of this package:
`% ls -la Payload/                                                                          
total 0
drwxr-xr-x  4 marg  staff  128 Oct  6 17:26 .
drwxr-xr-x  6 marg  staff  192 Oct  6 17:26 ..
drwxr-xr-x  3 marg  staff   96 Sep 28 19:08 Applications
drwxr-xr-x  5 marg  staff  160 Sep 28 19:08 Library`

In an AutoPkg recipe, one can now immediately use the results for repackaging.

homebysix:
I'm interested in incorporating this — great idea! However if we merge as is, it would result in the breakage of any recipe that expects the Payload file (not directory) to exist for subsequent unpacking.

As an example, here's a snippet of output from a Google Earth recipe that uses FlatPkgUnpacker modified to include `--expand-full`:

```
[...]
FlatPkgUnpacker
{'Input': {'destination_path': '~/Library/AutoPkg/Cache/com.github.ahousseini-recipes.pkg.GoogleEarthPro/unpack',
           'flat_pkg_path': '~/Library/AutoPkg/Cache/com.github.ahousseini-recipes.pkg.GoogleEarthPro/downloads/Google '
                            'Earth Pro.dmg/*Google Earth Pro*.pkg'}}
FlatPkgUnpacker: Mounted disk image ~/Library/AutoPkg/Cache/com.github.ahousseini-recipes.pkg.GoogleEarthPro/downloads/Google Earth Pro.dmg
FlatPkgUnpacker: Unpacked /private/tmp/dmg.NgbqzB/Install Google Earth Pro 7.3.6.10201.pkg to ~/Library/AutoPkg/Cache/com.github.ahousseini-recipes.pkg.GoogleEarthPro/unpack
{'Output': {}}
PkgPayloadUnpacker
{'Input': {'destination_path': '~/Library/AutoPkg/Cache/com.github.ahousseini-recipes.pkg.GoogleEarthPro/version_check',
           'pkg_payload_path': '~/Library/AutoPkg/Cache/com.github.ahousseini-recipes.pkg.GoogleEarthPro/unpack/Google_Earth_Pro.pkg/Payload'}}
extraction of ~/Library/AutoPkg/Cache/com.github.ahousseini-recipes.pkg.GoogleEarthPro/unpack/Google_Earth_Pro.pkg/Payload with aa failed
Failed.
Receipt written to ~/Library/AutoPkg/Cache/com.github.ahousseini-recipes.pkg.GoogleEarthPro/receipts/GoogleEarthPro.pkg-receipt-20250426-125935.plist

The following recipes failed:
    ~/Developer/repo-lasso/repos/autopkg/ahousseini-recipes/GoogleEarthPro/GoogleEarthPro.pkg.recipe
        Error in com.github.ahousseini-recipes.pkg.GoogleEarthPro: Processor: PkgPayloadUnpacker: Error: extraction of ~/Library/AutoPkg/Cache/com.github.ahousseini-recipes.pkg.GoogleEarthPro/unpack/Google_Earth_Pro.pkg/Payload with aa failed
```

The best way to avoid this might be to leverage the MinimumVersion information in recipes, meant to specify which version of AutoPkg the recipe requires. If the MinimumVersion of the recipe is set to {future version that includes `--expand-full`} then we go ahead and do `--expand-full`, otherwise we default to good ol' `--expand`. That's quite a bit more coding, but worth it to maintain backwards compatibility I think.

PR #914, opened 2023-11-14
Refactor to use poetry

hurricanehrndz:
I wanted to get some feedback on this PR before I continue any further, just in case there is not interest in these changes. They could potentially be breaking.

So currently autopkg is not package in a very Pythonic way. In fact the current layout makes it very hard to distribute as a Python package. I would like to address this.

So I have started restructuring autopkg in order to achieve that goal. At the same time I have been refactoring the project to support Python 3.11 and 3.12

I have packaging left, and fix loading of processors found in recipe directories. Anyhow, thoughts and feedback welcomed.


nmcspadden:
I wish  you had reached out to us before starting down this path, as I would have generally dissuaded you from doing so without consulting us on the direction.

First and foremost - all development is done on the dev branches, not master, so master is not representative of where the codebase is currently. Hence, a good amount of the work you've done will need to be rebased and checked against dev, and also against the new dev branch where I am refactoring the recipe logic.

I certainly appreciate the desire to help here, but I will also strongly discourage you from filing giant PRs like this - it's practically impossible to effectively review since you've mostly rewritten the codebase. Instead, I encourage you to break things up into very specific problems that are easily tested, compared to existing procedures, and measured for effectiveness. 

If you wish to contribute further, and I certainly hope you do, you can start by looking at the dev branch and the dev new recipe branches and make sure you start small, one section or problem space at a time. Feel free to come discuss in #autopkg on Slack, or on the mailing list. 

But as is, I can't ever accept a PR like this.

hurricanehrndz:
Thank you so much for the feedback. I am more than willing to rebase all my changes on dev.

I can definitely break them down. Did you wish me to fix the test as I migrate the code to poetry?

hurricanehrndz:
Also they look like a lot of change, but they are really not.

PR #915, opened 2023-11-14
Create python wheel

hurricanehrndz:
Make it possible to produce a python wheel form this project.

Benefits:
* improved installation experience on non Darwin platforms
    ```sh
    pip install git+https://github.com/autopkg/autopkg.git@dev
    ```
* adheres more closely to PEP-338
* improved developer experience for testing:
   ```sh
   python -mvenv venv
   source venv/bin/activate
   pip install -r gh_actions_requirements.txt
   python -m build
   pip install --force dist/*.whl
   rehash
   autopkg version
   ```

Known deficiencies:
* version number in pyproject.toml is hardcoded
* wheel more than likely will not be accepted into python.org pypi server due to packages in this repo all being top-level

### DIFF
--- a/Code/tests/test_expand_repo.py
+++ b/Code/tests/test_expand_repo.py
@@ -12,13 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import imp
-import os
 import unittest
 
-autopkg = imp.load_source(
-    "autopkg", os.path.join(os.path.dirname(__file__), "..", "autopkg")
-)
+from autopkgcli import autopkg
 
 
 class TestExpandRepo(unittest.TestCase):
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,29 @@
+[build-system]
+requires = ["setuptools", "setuptools-scm"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "autopkg"
+authors = []
+version = "2.7.3"
+description = "AutoPkg is an automation framework for macOS software packaging and distribution"
+readme = "Readme.md"
+requires-python = ">=3.10"
+license = {text = "BSD-3-Clause"}
+classifiers = [
+    "Topic :: System :: Software Distribution",
+]
+dynamic = [ "dependencies" ]
+
+[project.scripts]
+autopkg = "autopkgcli.autopkg:main"
+
+[tool.setuptools.dynamic]
+dependencies = {file = ["gh_actions_requirements.txt"]}
+
+
+[tool.setuptools.packages.find]
+where = ["Code"]
+include = ["autopkg*", "nuget*"]
+exclude = [".tests*"]
+namespaces = false
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,166 @@ autopkg.code-workspace
 # don't track package builds
 Scripts/artifacts/*
 Scripts/pkg-scripts/*
+
+### Python
+# Byte-compiled / optimized / DLL files
+__pycache__/
+*.py[cod]
+*$py.class
+
+# C extensions
+*.so
+
+# Distribution / packaging
+.Python
+build/
+develop-eggs/
+dist/
+downloads/
+eggs/
+.eggs/
+lib/
+lib64/
+parts/
+sdist/
+var/
+wheels/
+share/python-wheels/
+*.egg-info/
+.installed.cfg
+*.egg
+MANIFEST
+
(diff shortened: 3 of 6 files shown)


pyproject.toml
```diff
@@ -0,0 +1,29 @@
+[build-system]
+requires = ["setuptools", "setuptools-scm"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "autopkg"
+authors = []
+version = "2.7.3"
```
jgstew, line 8: can this be sourced from another file? I know it is technically possible, but would be nice if the version was single sourced.

pyproject.toml
```diff
@@ -0,0 +1,29 @@
...
+classifiers = [
+    "Topic :: System :: Software Distribution",
+]
+dynamic = [ "dependencies" ]
+
+[project.scripts]
+autopkg = "autopkgcli.autopkg:main"
+
+[tool.setuptools.dynamic]
+dependencies = {file = ["gh_actions_requirements.txt"]}
```
jgstew, line 22: I agree with this, the other dependency files have too much stuff in them.

jgstew:
The biggest change this incurs is moving the main code from `autopkg` to `autopkgcli/autopkg.py` which is required to make this work best, but also a potentially breaking change that I don't personally know how to test sufficiently on MacOS since that isn't my area.

Code/tests/test_expand_repo.py
```diff
@@ -12,13 +12,9 @@
...
 # limitations under the License.
 
-import imp
-import os
 import unittest
 
-autopkg = imp.load_source(
-    "autopkg", os.path.join(os.path.dirname(__file__), "..", "autopkg")
-)
+from autopkgcli import autopkg
```
jgstew, line 17: Same issue: I think this will have an issue if you are trying to dev/test autopkg source, then this will actually load autopkg from the pip installed version, not from the source version you are currently testing as a cloned repo.

I think this should only be a problem if you have previously pip installed autopkg? or maybe it only works if you have pip installed autopkg and won't work otherwise? I don't remember exactly.

jgstew:
@nmcspadden  I think you have to approve this PR to run the github action automated tests.

pyproject.toml
hurricanehrndz: Version is in plist which is incompatible with setuptools

pyproject.toml
hurricanehrndz: There is a lot in the current requirements that could be qualified as dev tools and not actually needed for the running of app. This should be trimmed.

Code/tests/test_expand_repo.py
hurricanehrndz: Not if you start using env python as your dev python

hurricanehrndz: If this get merged, additional PRs will obviously need to be done to update docs on install and dev environment setup.

pyproject.toml
jgstew: where is the plist that contains the version? I was looking for it.

Code/tests/test_expand_repo.py
jgstew: well I don't think using this method is required especially for the default use case on MacOS in which case the PKG would still be preferred.

jgstew:
The unit test failed due to:

github.GithubException.RateLimitExceededException: 403 {"message": "API rate limit exceeded for 13.105.49.65. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)"

it seems like the github token isn't being read from the default location if no default location is specified, which then causes this issue. This is due to a different bug.

jgstew:
I do think the fact that this change moves the core autopkg code from one file to another is a pretty significant change, and one that might not make sense at the moment with the potential impact it could have.

I think this would be an impediment to getting autopkg 3.x out the door as the main release version, which I think is higher priority than the potential problems this change could introduce at the moment.

I do like this concept in general though and do think it is worth future consideration.

pyproject.toml
jgstew: I do agree, it makes sense to make a distinction between requirements for run/build vs requirements for dev/test/validation. I do think there is too much stuff in some of the requirements files for the sake of doing the basics in a github action or just running recipes on a build server.

I think the reason for the additional items in the requirements is because the requirements were traditionally only ever used by people doing dev work, while the production way to use autopkg was to install the PKG file on MacOS.

PR #925, opened 2024-01-24
Add curl_common_opts to prefetech_filename

n8felton:
This allows for request_headers to be added to the curl command so that headers such as `Authorization` will be included while attempting to prefetch the filename, otherwise the prefetch will fail.


homebysix (approved):

homebysix:
Thanks!

PR #926, opened 2024-01-25
Dev new recipe logic

nmcspadden:


Code/autopkglib/recipes/__init__.py
```diff
@@ -0,0 +1,762 @@
...
+import pathlib
+import plistlib
+import pprint
+import sys
+from dataclasses import dataclass, field
+from typing import Any, Dict, List, Optional
+
+import yaml
+
+sys.path.append("/Users/nmcspadden/Documents/GitHub/autopkg/Code")
```
gregneagle, line 29: Whatever this was a quick-and-dirty workaround for; you should remove it now.

Code/autopkglib/recipes/__init__.py
```diff
@@ -0,0 +1,762 @@
...
+    # recipe.display_chain()
+    recipe = fetch_recipe_chain("Firefox.munki", check_only=False)
+    recipe.display_chain()
+    # print("** Dictionary version")
+    # rdict = recipe.to_dict()
+    # pprint.pprint(rdict, width=1)
+    # print("** Checking non-git recipe")
+    # non_git_recipe = "/Users/nmcspadden/Library/AutoPkg/Recipes/AutoPkg-Test.download.recipe"
+    # hash = get_git_commit_hash(non_git_recipe)
+    # print(f"Git hash: {hash}")
```
n8felton, line 762: Not sure how much of this you want to keep in the final PR.

Code/autopkglib/recipes/__init__.py
```diff
@@ -0,0 +1,762 @@
...
+        }
+
+    def verify_trust(self) -> bool:
+        """Return True if the recipe trust is correct."""
+        return True
+        # TODO: Implement this
+        # We need to determine if a recipe in the chain is an override and thus contains trust
+        # if it contains trust, we then go validate that the trust is correct
+        # if there are no overrides, this always returns True (but maybe we print out that we did nothing)
+        # for recipe in self.recipes:
```
n8felton, line 257: Should this be implemented before this PR is merged?

Code/autopkglib/recipes/__init__.py
```diff
@@ -0,0 +1,762 @@
...
+
+import glob
+import json
+import os
+import pathlib
+import plistlib
+import pprint
+import sys
+from dataclasses import dataclass, field
+from typing import Any, Dict, List, Optional
```
n8felton, line 25: I know that `typing` is littered all around the code right now, but maybe AutoPkg 3.0 should be the start of the transition to generic types, considering v2.6+ was built using Python 3.10.

I'm not saying go through _all_ the code to refactor right now, but maybe start with some of the net-new stuff.

References:
- https://stackoverflow.com/a/62033243/2233241
- https://stackoverflow.com/a/77300462/2233241

Code/autopkglib/recipes/__init__.py
nmcspadden: Removed! Thankfully nothing broke...

Code/autopkglib/recipes/__init__.py
nmcspadden: This should be fixed now! Thanks for the advice.

Code/autopkglib/recipes/__init__.py
nmcspadden: Yeah this will all come out. Was part of the testing process.

homebysix:
Updated `dev` and `dev_new_recipe_logic` branches from master to reflect 2.7.3 release.

homebysix:
This work conflicts with the newer work in #1027 now merged into the main branch and intended to be the basis for a future 3.0 release.

There's potentially some good ideas captured here that we should consider in the future, but they'd need to be retargeted at the new dev branch. I don't think any of them are must-haves for 3.0, so I'm closing this out.

PR #929, opened 2024-01-30
YAML - On load, convert all scalar types to strings

MLBZ521:
This change will use the YAML SafeConstructor to safely load YAML data and auto-convert scalar types to strings to provide a consistent experience with AutoPkg's current behavior and data type expectations.

Basically this change will perform a very similar result as what I proposed in #911, however, it will only convert scalar types while preserving non-scalar types.  Doing it this way should prevent any issues with current recipes.

I think, short term, this is a better option that what I proposed in #911.

Long term, I think updating the [do_variable_substitution](https://github.com/autopkg/autopkg/blob/142368f84344186cbe3105d94e329e3d901da157/Code/autopkglib/__init__.py#L466C2-L486) function (or [update_data](https://github.com/autopkg/autopkg/blob/142368f84344186cbe3105d94e329e3d901da157/Code/autopkglib/__init__.py#L458-L488) function as whole) would create the best experience as suggested by @jgstew [here](https://github.com/autopkg/autopkg/issues/895#issuecomment-1906119854).  This would give recipe writers more flexibility and allow more data types to be passed through AutoPkg steps.


homebysix:
This should be superseded by https://github.com/autopkg/autopkg/pull/1023, which is now merged into the dev branch, which will be the basis of a future 3.0 release.

One important difference between https://github.com/autopkg/autopkg/pull/1023 and this PR is that https://github.com/autopkg/autopkg/pull/1023 preserves pure integer types, which preserves compatibility with custom processors that require them (such as [VersionSplitter](https://github.com/autopkg/homebysix-recipes/blob/master/VersionSplitter/VersionSplitter.py)).

PR #936, opened 2024-02-14
Actually bypass when building map

n8felton:


homebysix:
This change is superseded by PR #1027, which is now in the dev branch and will be the basis of a future 3.0 release. Thanks for submitting!

PR #953, opened 2024-08-16
Changes to allow running custom processors via command line usage/arguments

macmule:
This PR contains 3 changes to `autopkglib/__init__.py` which were found to be needed in order to run a customer processor with command line usage/arguments (as per: [https://github.com/autopkg/autopkg/wiki/Developing-Custom-Processors#sample-usage---command-line-argumentsvariables](https://github.com/autopkg/autopkg/wiki/Developing-Custom-Processors#sample-usage---command-line-argumentsvariables)).

I also raised a discussion on this: [https://github.com/autopkg/autopkg/discussions/952](https://github.com/autopkg/autopkg/discussions/952)

Basically, running a custom processors with a plist works.. but found that running custom processors via command line usage/arguments did not.. and found another issue too.

I'm happy though to have this PR closed without merging, if it's decided to just drop support for custom processors via command line usage/arguments. I can then submit another PR to remove from `autopkglib/__init__.py` and update: [https://github.com/autopkg/autopkg/wiki/Developing-Custom-Processors](https://github.com/autopkg/autopkg/wiki/Developing-Custom-Processors)

The changes are in their own commits, I've linked and added the commit messages below:

- 6675343 -  Parses self.env, and creates a new dict only with keys whose value is not None.
- 9e37ddf - Resolves an "TypeError: 'NoneType' object does not support item assignment" issue when calling a custom processor via command line usage/arguments.. as self.env is empty.. so this substantiates self.env as an empty dict.. which can then be appended to.
- e9db7b2 - Resolves an issue where autopkg would appear to hang when trying to call a custom processor via command line usage/arguments.

### DIFF
--- a/Code/autopkglib/__init__.py
+++ b/Code/autopkglib/__init__.py
@@ -585,11 +585,17 @@ class Processor:
         if self.env is None:
             return
 
+        plist_safe = {}
+        
+        for env_key in self.env:
+            if not self.env[env_key] is None:
+                plist_safe[env_key] = self.env[env_key] 
+
         try:
             with open(self.outfile, "wb") as f:
-                plistlib.dump(self.env, f)
+                plistlib.dump(plist_safe, f)
         except TypeError:
-            plistlib.dump(self.env, self.outfile.buffer)
+            plistlib.dump(plist_safe, self.outfile.buffer)
         except BaseException as err:
             raise ProcessorError(err) from err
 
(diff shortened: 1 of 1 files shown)


gregneagle:
I think making it easier to develop and debug custom processors is always a good thing, and I support getting stuff working again the way it did in the past.

PR #959, opened 2024-11-16
feat: add S3Downloader processor for S3 role based downloads

dennishenry:
In working with my company's Business Technology team, it came to my attention that `autopkg` did not support role based cloning, so I wrote up this processor to fill the need, and wanted to contribute it upstream. It appears to work without issue but I am new to the codebase and tool so I don't know what I don't know 😅 Thanks in advance for any and all thoughts!

### DIFF
--- a/Code/autopkglib/S3Downloader.py
+++ b/Code/autopkglib/S3Downloader.py
@@ -1,0 +1,91 @@
+#!/usr/local/autopkg/python
+#
+# Copyright 2024 Dennis Henry @dennishenry
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


dennishenry:
CC @homebysix as what seems to be the most active contributor

homebysix:
I would suggest hosting this processor in your own repo (or the repo of an involved collaborator in the autopkg org) instead of adding it to AutoPkg's list of core processors.

Rationale: Adding a requirement for the boto3 module will increase the complexity of packaging and testing the Python framework included with AutoPkg, and such effort would only benefit a few people in the community who leverage this processor. Better to leave the testing, documentation, and support of this processor to the people who actually use it regularly.

Some examples of processors not included in AutoPkg's core, some of which require nonstandard Python modules, can be found on the [Noteworthy Processors](https://github.com/autopkg/autopkg/wiki/Noteworthy-Processors) page. Feel free to add a link to this processor to that page, once you have a repository set up for it.

dennishenry:
Makes perfect sense @homebysix! Will do and modify this PR to be a link. Do you think this would be useful for others?

Issue #934, opened 2024-02-08
GitHubReleasesInfoProvider additional functionality - raise a warning if a repo has been archived or disabled. 


paul-cossey:
**Describe the problem**

More a feature request than a problem... raising following a conversation in [MacAdmins Slack](https://macadmins.slack.com/archives/C056155B4/p1707211066391709)

 `GitHubReleasesInfoProvider` does not raise a warning or error if a repo has been archived or disabled. If the media is still available, recipes will complete successfully, and admins have no way of knowing the project state unless there is a manual check. 

**Preferences contents**
N/A

**AutoPkg output**
N/A

**Expected behaviour**

If `GitHubReleasesInfoProvider` detects that a repo is either `"archived": true,` or  `"disabled": true,` it should raise a warning to inform the admin that investigation is need. 

**Version (please complete the following information):**
 - OS version: N/A
- AutoPkg Version: N/A

homebysix:
@nmcspadden I would suggest the opposite: generate a warning by default. This would treat archived GitHub repos similarly to deprecated recipes (with DeprecationWarning) — equivalent situations in which future recipe runs are unlikely to produce useful results for AutoPkg users, but which also cause no direct harm.

homebysix:
Diving deeper into this, I think both options (warning or error) are worse than the current state.

- The "archived" attribute isn't surfaced as part of GitHubReleasesInfoProvider's API call to the releases endpoint; that'd need a separate call to the repo object. For AutoPkg users not using a GitHub token, this would double the requests eligible for rate limiting, for every release check on every repo, whether actually archived or not. We could work around that by making the archived check optional, but:

- Archival is reversible. A repo becoming unarchived and subsequently posting a new release is probably _super_ rare, but not impossible.

- Archived repo assets are still available. Much like developers sometimes host software that is no longer updated but still downloadable (like Rogue Amoeba's [Legacy Software](https://rogueamoeba.com/legacy/)), it may be useful or familiar to AutoPkg users to run a recipe expecting to get the last published version. It would be odd if URLTextSearcher showed an error if it sees the string "discontinued" on the page, or if URLDownloader sees that the `last-modified` header is many years ago, right?

- Disabled repos are a clearer signal, but I suspect trying to use GitHubReleasesInfoProvider with a disabled repo already results in a 401 or 451 code and subsequent ProcessorError. Please correct me if you have firsthand info on that.

[...]

PR #961, opened 2025-01-05
fail autopkg hard if GitHubReleasesInfoProvider.py queries an archived repo, unless fail_if_archived is False

discentem:


Code/tests/test_githubreleasesinfoprovider.py
```diff
@@ -74,7 +76,28 @@ def test_returns_asset_url(self):
         self.processor.env = test_env
         self.processor.main()
         self.assertIsNotNone(test_env["asset_url"])
+    
+    @patch('autopkglib.GitHubReleasesInfoProvider.is_archived')
```
discentem, line 80: Please let me know if I should add more tests elsewhere

discentem:
Not sure why `TestGithubReleaseInfoProvider (macos-latest)` is failing 😞

homebysix:
I like this concept, but I have a nitpick about this mismatch between the input variable name (`ignore_archived`) and its behavior (raises an exception, rather than "ignoring"). I don't think it's desirable for an exception to be raised, since even archived repositories can successfully provide software download assets.

I could envision something similar to DeprecationWarning, where GitHubReleasesInfoProvider raises a warning in the recipe summary output to notify the AutoPkg user that the GitHub repo is archived, like so:

```
The following recipes raised GitHub release warnings:
    Name          Warning
    ----          -------
    Foo.download  The GitHub repository hosting assets for this recipe has been
                  archived, and is therefore unlikely to provide future updates.
```

Unfortunately there's no common "warnings" summary output for AutoPkg; it's up to each processor to determine what it adds to its `*_summary_result` variables. So this would be a new `github_releases_summary_result` output variable, probably.

discentem:
Thanks for the reply @homebysix! I implemented it this way due to Nick's comment here: https://github.com/autopkg/autopkg/issues/934#issuecomment-1934576200

However, I'm happy to change the variable name, reverse the behavior, and/or switch to a warning instead of ProcessorError if that is the preferred direction.

discentem:
Oh, I definitely made a mistake with my current implementation. But only in the description I think?

1. The description is wrong:

```diff
"description": (
- "If set to True, the processor will raise an exception if the "
-  "GitHub repo is archived."
+  "If set to False, the processor will raise an exception if the Github repo is archived."
),
```

nmcspadden:
Is this something you're still interested in fixing?

discentem:
> Is this something you're still interested in fixing?

Yeah, sure thing! Other than updating the description, are there any changes you want to see? Also if you have any advice about the failing test that would be very helpful.

PR #962, opened 2025-02-04
Updating to pull cert bundle from SSL_CERT_FILE if set

smithjw:
This PR adds a small change into `ssl_context_certifi(self)` that will use the certificate bundle defined in the `SSL_CERT_FILE` environment variable if set.

It's common for SSL Inspection to be used on corp networks and this allows users to use this processor without needing to disable SSL validation with the `disable_ssl_validation` input.

### DIFF
--- a/Code/autopkglib/URLDownloaderPython.py
+++ b/Code/autopkglib/URLDownloaderPython.py
@@ -331,10 +331,15 @@ class URLDownloaderPython(URLDownloader):
         return ctx
 
     def ssl_context_certifi(self):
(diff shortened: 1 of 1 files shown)


homebysix:
@smithjw - Just verifying you've tested this internally and it works for you? I don't have a custom cert bundle to test with.

Feel free to ignore copilot if you don't like its suggestion.

homebysix:
And this has already been in use in your environment successfully? If so I'll aim to get this into 2.7.6.

Issue #966, opened 2025-05-12
Munki recipes/core processors unable to handle variables for timestamp/date types


arubdesu:
With a clean 2.7.4 setup and just the preferences required to point at a local Munki repo (and the [DatetimeProcessor](https://github.com/autopkg/haircut-recipes/blob/master/Processors/DatetimeOutputter.py) from haircut-recipes), an autopkg recipe (or override) cannot successfully create a `<date>` type (in XML, or !!timestamp in YAML) e.g. for setting the `force_install_after_date`.
Trying to set it manually in the override causes plutil to fail validation since it sees the variable as a string in a <date> value, `Could not interpret <date>`, autopkg/python similarly gives a traceback if you went ahead anyway.

Here's a snippet of the recipe in the 'polluted override' method I've used as a hack for a good long while, tacked to the end of (as a small example) VLC.download.recipe's override (here's the full [zip](https://github.com/user-attachments/files/20156191/VLC.download.recipe.zip))
:
```
    <key>Process</key>
    <array>
        <dict>
            <key>Processor</key>
            <string>com.github.haircut.processors/DatetimeOutputter</string>
        </dict>
        <dict>
            <key>Processor</key>
            <string>MunkiImporter</string>
            <key>Arguments</key>
            <dict>
                <key>pkg_path</key>
                <string>%pathname%</string>
                <key>repo_subdirectory</key>
                <string>%MUNKI_REPO_SUBDIR%</string>
                <key>pkginfo</key>
                <dict>
                    <key>force_install_after_date</key>
                    <string>%datetime%</string>
```

**Expected behavior**
While the DatetimeProcessor only currently outputs a string, with expanded support of an additional flag to pass the datetime as-is (preserving the Jamf-focused original behavior), and the ability to detect its use in the applicable Munki-specific core processors, it **should** result in a proper yyyy-mm-ddThh:mm:ssZ-formatted timestamp with date as the type in the raw XML of the pkginfo. Instead, currently the core MunkiImporter processor will end up with this as a result of the override pasted above:
```
% grep -A 1 force /Users/Shared/munki_repo/pkgsinfo/apps/VLC/VLC-3.0.21.plist
	<key>force_install_after_date</key>
	<string>2025-05-12T15:48:25.889619</string> 
```
A side-effect of setting the `force_install_after_date` to a string is that MunkiAdmin completely ignores the file and therefore doesn't render the pkg at all.

**Version (please complete the following information):**
 - macOS 15.3.1
[...]

PR #967, opened 2025-05-13
Special-case force_install_after_date in pkginfo

arubdesu:
This change enables MunkiImporter to create a working <date> type when the pkginfo key being read in is for force_install_after_date, as that's the only key of this type and reading in the serialized plist with a date key set to a variable wouldn't work. Handles if you leave the Z on the end of the datetime format (or otherwise screw it up)

### DIFF
--- a/Code/autopkglib/MunkiImporter.py
+++ b/Code/autopkglib/MunkiImporter.py
@@ -18,6 +18,7 @@
 import os
 import plistlib
 import subprocess
+from datetime import datetime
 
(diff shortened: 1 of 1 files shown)


arubdesu:
Should address #966 if I done did it right, tested with a few of my browser 'overloaded overrides' recipes

nmcspadden:
Please move this to autopkg:dev, but otherwise I'm okay with this.

Issue #899, opened 2023-09-27
AutoPkg does not read GITHUB TOKEN from default path on fresh setup


jgstew:
**THIS IS ONLY INTENDED FOR AUTOPKG BETAS.**

**Describe the problem**

When running AutoPkg with a blank preference file, I expect the GitHub token to be read from the default path, but it is not.

**Preferences contents**
*BE SURE TO SANITIZE ANY SENSITIVE DATA SUCH AS PASSWORDS OR ADDRESSES.*

N/A

**AutoPkg output**
*BE SURE TO SANITIZE ANY SENSITIVE DATA SUCH AS PASSWORDS OR ADDRESSES.*

See GitHub actions in PR

**Expected behavior**
A clear and concise description of what you expected to happen. What specific part of the recipe or AutoPkg run did not behave correctly?

**Version (please complete the following information):**
 - OS version: all
- AutoPkg Version: 3.0.0RC2

homebysix:
This should be fixed on the dev branch, which will be the basis of a future 3.0 release.

PR #972, opened 2025-06-20
Update default token path for GitHub access in AutoPkg v3

ndhillon-openai:
_This is for dev-3.x_

This PR addresses the following issue:

https://github.com/autopkg/autopkg/issues/899

Changes include:

- Updated default token path for GitHub access in AutoPkg.
- Added compatibility for Python 3.12 by replacing 'imp' module with 'importlib'. 
- Provided fallback implementation for 'appdirs'.

### DIFF
--- a/Code/autopkglib/__init__.py
+++ b/Code/autopkglib/__init__.py
@@ -16,7 +16,38 @@
 
 """Core/shared autopkglib functions"""
 import glob
-import imp
+# The `imp` module was removed in Python 3.12. AutoPkg still relies on the
+# `imp.load_source()` helper to dynamically load processors during runtime and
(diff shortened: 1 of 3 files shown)


ndhillon-openai:
Hi @homebysix  @gregneagle — whenever you have a moment, could you please take a look?
This PR targets dev-3.x and unblocks Python 3.12 compatibility.

gregneagle:
This feels a lot like 2 or 3 unrelated things combined into a single PR, making it hard to review and test. Are these changes dependent on each other?

nmcspadden:
I agree - this needs to be separated a bit so that each change is a separate PR. As is, I'm not willing to merge like this.

PR #976, opened 2025-07-12
Allow custom output variable name for FindAndReplace Processor

jgstew:
Add ability to set custom output variable name.

I was already doing basically the same thing in some of my other processors, so I basically lift and shifted the code to this one.

Example output from a [test recipe](https://github.com/jgstew/jgstew-recipes/blob/main/Recipes-Examples/FindAndReplace.test.recipe.yaml):

```
Processing com.github.jgstew.test.FindAndReplace...
WARNING: com.github.jgstew.test.FindAndReplace is missing trust info and FAIL_RECIPES_WITHOUT_TRUST_INFO is not set. Proceeding...
FindAndReplace
{'Input': {'find': '_',
           'input_string': '1_2_3',
           'replace': '.',
           'replace_output_var': 'version'}}
FindAndReplace: Replacing "_" with "." in "1_2_3".
{'Output': {'version': '1.2.3'}}
com.github.jgstew.SharedProcessors/AssertInputContainsString
{'Input': {'assert_string': '1.2.3', 'input_string': '1.2.3'}}
AssertInputContainsString: No value supplied for raise_error, setting default value of: True
AssertInputContainsString: `1.2.3` must contain `1.2.3`
{'Output': {'assert_result': 'found!'}}
```

### DIFF
--- a/Code/autopkglib/FindAndReplace.py
+++ b/Code/autopkglib/FindAndReplace.py
@@ -44,6 +44,11 @@ class FindAndReplace(Processor):
             "description": 'The string that you want to replace the "find" '
             "string with.",
         },
+        "replace_output_var": {
+            "required": False,
+            "default": "output_string",
+            "description": "Variable to store the output to, defaults to `output_string`",
+        },
     }
     output_variables = {
         "output_string": {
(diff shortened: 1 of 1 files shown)


Code/autopkglib/FindAndReplace.py
```diff
@@ -44,6 +44,11 @@ class FindAndReplace(Processor):
             "description": 'The string that you want to replace the "find" '
             "string with.",
         },
+        "replace_output_var": {
```
jgstew, line 47: not sure if there is a better name to use than `replace_output_var` for this.

homebysix:
Thanks! I might adjust the key to `result_output_var_name` to match URLTextSearcher.

PR #990, opened 2025-10-13
Update to Python 3.10.11

MScottBlake:
This pull request changes the various python references to 3.10.11, the latest version of 3.10 with binary installers.

I ran the `Scripts/run_tests.py` file using 3.10.11. These are the results:
```
Ran 708 tests in 2.285s

OK (skipped=12)
```

The only questionable thing I did in this PR is that I also updated `.github/workflows/requirements.txt` to match `requirements.txt`. These two files were very similar already, so I think it's safe, but I don't have the context why there are different files in the first place. I'm happy to undo or modify that change as necessary.

### DIFF
--- a/Code/autopkglib/__init__.py
+++ b/Code/autopkglib/__init__.py
@@ -17,6 +17,7 @@
 """Core/shared autopkglib functions"""
 import glob
 import imp
+import importlib.resources
 import json
 import os
 import plistlib
@@ -30,7 +31,6 @@ from distutils.version import LooseVersion
 from typing import IO, Any, Union
 
 import appdirs
-import pkg_resources
 import yaml
 
 # Type for methods that accept either a filesystem path or a file-like object.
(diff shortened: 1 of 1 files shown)


homebysix:
This required quite a bit of tinkering, but I finally got an AutoPkg release dry-run to succeed on this branch. My goal is to merge this in to dev-2.x for a future 2.9 release, however I need to run some tests for a while to ensure there are no regressions.

I adjusted the make-release script to bypass any conflicts in the local Python environment by temporarily setting aside user site-packages and pip ([8ef3774](https://github.com/autopkg/autopkg/pull/990/commits/8ef377463935f5de11e7f0a395b37a2db630320e) and [da9bb1b](https://github.com/autopkg/autopkg/pull/990/commits/da9bb1b35da2a92933df4949358f055454217dc0)).

I also had to update from `pkg_resources` to `importlib.resources` to prevent this warning from appearing when running any verb ([f888e4c](https://github.com/autopkg/autopkg/pull/990/commits/f888e4ce67cd46987930ee78d49f4b12bdb5aa44)):

```
% autopkg version
/Library/AutoPkg/autopkglib/__init__.py:33: UserWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html. The pkg_resources package is slated for removal as early as 2025-11-30. Refrain from using this package or pin to Setuptools<81.
  import pkg_resources
2.7.7
```

I'll continue testing and tweaking and update this thread.

homebysix:
I've been testing this in combination with the new cached search index changes in the dev-2.x branch, and it seems to be working well.

Did you do a global bulk `pip install --upgrade` for all the requirements, or did you choose specific modules/versions that align with security vulnerability fixes or features AutoPkg uses?

PR #1024, opened 2026-04-12
Add global option for recipe override format

grahampugh:
This pull request updates the `make_override` function in `Code/autopkg`. The main change is to allow the override format to be set via a global preference, with explicit command-line arguments still taking precedence. 

**Override format resolution and usage:**

* The `--format` CLI argument now defaults to `None` instead of `"plist"`, and the help text is updated to mention that the format can be set globally via the `RECIPE_OVERRIDE_FORMAT` preference (this is case-insensitive, so `recipe_override_format` will also work).
* The override format is now resolved in priority order: command-line argument > global preference > default (`"plist"`). The resolved format is stored in the `override_format` variable and used throughout the function.
* All subsequent checks and file naming logic now use `override_format` instead of directly referencing `options.format`, ensuring consistent behavior regardless of how the format is set.

### DIFF
--- a/Code/autopkg
+++ b/Code/autopkg
@@ -1855,16 +1855,22 @@ def make_override(argv):
     parser.add_option(
         "--format",
         action="store",
-        default="plist",
+        default=None,
         help=(
             "The format of the recipe override to be created. "
-            "Valid options include: 'plist' (default) or 'yaml'"
+            "Valid options include: 'plist' (default) or 'yaml'. "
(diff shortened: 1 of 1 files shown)


homebysix (approved):

homebysix:
Thanks! We'll need to update the wiki prior to next 2.x release.

grahampugh:
> Thanks! We'll need to update the wiki prior to next 2.x release.

Happy to do that now, if it wouldn't cause confusion in the meantime? Or, let me know when you're ready to make a release and I'll do it then.

PR #1026, opened 2026-04-20
Add faster parsing of recipe overrides utilising ParentRecipeTrustInfo

smithjw:
## Summary

Adds an fast path for locating and loading overrides that have `ParentRecipeTrustInfo`. The existing path is untouched and is still used whenever the fast path does not apply or does not succeed.

## Motivation
`find_recipe_by_identifier` globs every search directory and parses every matching recipe file to read its identifier. `load_recipe` recurses through the parent chain, which triggers a full scan per recipe parent. For a three-parent recipe chain with a few hundred overrides and a few thousand recipes across installed repos, this takes several seconds of pure filesystem/parsing work before a recipe actually starts running.

Overrides already carry `ParentRecipeTrustInfo.parent_recipes[*].path`, which records the absolute path of every parent recipe. This is enough information to skip the scan entirely for common CI cases like running overrides by name.

This is an attempt at speeding up recipe runs/loading while not incorporating a completely new feature like the Recipe Map

## What changes

In `Code/autopkg`:
- New `_fast_locate_override_by_name`: direct filename lookup in `override_dirs` using `os.scandir`
- New `_try_assemble_from_trust_info`: builds a merged recipe dict using the paths in `ParentRecipeTrustInfo`
- New `_merge_child_into_parent`: extracted helper mirroring the merge block in `load_recipe`, reused by the fast path to guarantee both paths produce structurally identical merged recipes
- New `_FAST_PATH_CACHE`: module-level dict used to hand the override's parsed dict from `_fast_locate_override_by_name` to `load_recipe` without re-parsing. Entries are consumed (popped) on first use
- Two-line change in `locate_recipe` to invoke the fast path
- Few-line change in `load_recipe` to consume `_FAST_PATH_CACHE` and to try `_try_assemble_from_trust_info` when trust info is present

## Performance

Measured on a setup with 9 recipe repos (~3300 recipes total), 319 overrides across 15 subdirectories, and a typical three-parent chain (download -> pkg -> jamf upload):

| Scenario | Baseline | Fast path | Speedup |
|---|---:|---:|---:|
| Single-recipe `autopkg info` (cold) | 7.78s | 0.34s | ~23x |
| Single-recipe `load_recipe` (warm)  | 7.57s | 0.007s | ~1000x |
| 10 recipes in one process           | 49.5s | 0.07s | ~700x |

## Tests

New `Code/tests/test_autopkg_fast_path.py` adds 34 tests across 5 classes:
- `TestFastLocateOverrideByName` (11 tests): path separator rejection, top-level and one-level-deep lookup, yaml and plist formats, name with/without recipe extension, multi-dir order, missing/invalid files, cache population
- `TestMergeChildIntoParent` (7 tests): identifier, input, process, minimum version, recipe path, parent recipes, description handling
- `TestTryAssembleFromTrustInfo` (8 tests): single- and three-level chains, preprocessors/postprocessors, missing files, identifier mismatch, broken chain, cycles, empty trust info
- `TestLocateRecipeFastPathIntegration` (6 tests): happy path, file path input bypass, identifier fall-through when no filename match, identifier input hits when override is filed under identifier, miss fall-through, plain recipe in repo still works
- `TestLoadRecipeFastPathEquivalence` (2 tests): golden structural equivalence between fast and slow paths; fall-back when a parent path becomes invalid

Full suite run before submitting.

### DIFF
--- a/Code/autopkg
+++ b/Code/autopkg
@@ -238,6 +238,199 @@ def get_repository_from_identifier(identifier: str):
     return correct_item["repository"]["name"]
 
 
+# Cache used by the override fast path. When _fast_locate_override_by_name
+# parses an override, it stashes the parsed dict here keyed by absolute
+# path so load_recipe can reuse later it rather than re-parsing the same file.
+# Entries are consumed (popped) by load_recipe.
+_FAST_PATH_CACHE: dict = {}
+
+
+def _merge_child_into_parent(parent, child, child_path=None):
+    """Mirrors the child-into-parent merge logic from load_recipe.
+
+    Used by _try_assemble_from_trust_info. Mutates ``parent`` in place and
+    returns it. Intentionally a straight port of the existing merge block
+    in load_recipe so both paths produce structurally identical results.
+    """
+    parent["Identifier"] = get_identifier(child)
+    parent["Description"] = child.get("Description", parent.get("Description", ""))
+    for key in list(child.get("Input", {}).keys()):
+        parent.setdefault("Input", {})[key] = child["Input"][key]
+
+    # take the highest of the two MinimumVersion keys, if they exist
+    for candidate in (parent, child):
+        if "MinimumVersion" not in candidate:
+            candidate["MinimumVersion"] = "0"
+    if version_equal_or_greater(child["MinimumVersion"], parent["MinimumVersion"]):
+        parent["MinimumVersion"] = child["MinimumVersion"]
+
+    parent["Process"] = parent.get("Process", []) + child.get("Process", [])
+
+    if parent.get("RECIPE_PATH"):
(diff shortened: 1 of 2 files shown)


homebysix:
Before I review, I wonder: why not just reimplement the recipe map feature from the 3.0RC branch? I asked in the #autopkg channel and feedback was generally positive about the recipe map performance improvements, so I'm inclined to go that direction.

smithjw:
@homebysix look, I did like the recipe map but the 3.x seemed like a very far future target that kept shifting 😅. I'll see if I can get it working in dev-2.x again

homebysix:
Thanks for #1027, which is now in the dev branch and should make this one unnecessary. Closing it out.

PR #1048, opened 2026-05-28
fix(PkgPayloadUnpacker): preserve ditto error when aa fallback also fails

### LINKED ISSUE #1047: PkgPayloadUnpacker: ditto error silently discarded when aa fallback also fails
**Describe the problem**

When `PkgPayloadUnpacker` fails to extract a payload with `ditto`, it falls back to `aa`. If both tools fail, the final `ProcessorError` only reports the `aa` failure — the `ditto` error (which contains the more informative diagnostic message) is silently discarded. Additionally, the `aa` error message does not include stderr output, making it impossible to understand why extraction failed from AutoPkg output alone.

There is also a latent `UnboundLocalError` bug: if `ditto` raises an `OSError` (i.e., the binary cannot be executed), `proc` is never assigned but is referenced unconditionally on the next line.

---

**Preferences contents**

N/A — this is a processor bug reproducible without any custom preferences.

---

**AutoPkg output**

```
PkgPayloadUnpacker
{'Input': {'destination_path': '/usr/local/autopkg/Cache/<recipe>/BeyondTrustJumpClient',
           'pkg_payload_path': '/usr/local/autopkg/Cache/<recipe>/unpack',
           'purge_destination': True}}
autopkglib.ProcessorError: extraction of /usr/local/autopkg/Cache/<recipe>/unpack with aa failed
```

With the fix applied, the actual error becomes visible:

```
autopkglib.ProcessorError: extraction of /usr/local/autopkg/Cache/<recipe>/unpack with ditto failed: ditto: /usr/local/autopkg/Cache/<recipe>/unpack: Is a directory
; extraction of /usr/local/autopkg/Cache/<recipe>/unpack with aa failed: Failed to create magic stream
Error creating input stream
extract failedError 0x9108c800
reading magic bytes
```

---

**Expected behavior**

When both `ditto` and `aa` fail, the `ProcessorError` should include the diagnostic output from **both** tools so the user can identify the root cause. In this case, the ditto error `"Is a directory"` immediately reveals that `pkg_payload_path` is pointed at the expanded package directory rather than the `Payload` file within it.

Specifically:
- The ditto stderr should not be overwritten when the `aa` fallback also fails.
- The `aa` error should include its stderr output (currently `proc.communicate()` discards the return value on line 111).
- `proc.returncode` on line 90 should be guarded against `UnboundLocalError` when `Popen` itself raises `OSError`.

---

**Proposed fix**

In `autopkglib/PkgPayloadUnpacker.py`, the `unpack_pkg_payload` method:

1. Rename `error` → `ditto_error` to prevent overwrite.
2. Guard `proc.returncode` with `if not ditto_error` to avoid `UnboundLocalError`.
3. Capture aa stderr: `_, aa_err_out = proc.communicate()`.
4. When both fail, raise with both messages combined.

---

**Version:**
- OS version: macOS 26.6 (25G5028f)
- AutoPkg version: 2.9.0

n8felton:
Fixes #1047

### DIFF
--- a/Code/autopkglib/PkgPayloadUnpacker.py
+++ b/Code/autopkglib/PkgPayloadUnpacker.py
@@ -70,7 +70,7 @@ class PkgPayloadUnpacker(Processor):
                     raise ProcessorError(f"Can't remove {path}: {err.strerror}")
 
         # set an error string when ditto or aa fail
-        error = ""
+        ditto_error = ""
         try:
             dittocmd = [
                 "/usr/bin/ditto",
@@ -84,15 +84,15 @@ class PkgPayloadUnpacker(Processor):
             )
             _, err_out = proc.communicate()
         except OSError as err:
-            error = (
+            ditto_error = (
                 f"ditto execution failed with error code {err.errno}: {err.strerror}"
             )
-        if proc.returncode != 0:
-            error = (
+        if not ditto_error and proc.returncode != 0:
+            ditto_error = (
                 f"extraction of {self.env['pkg_payload_path']} with ditto failed: "
                 f"{err_out}"
             )
-        if error and os.path.exists("/usr/bin/aa"):
+        if ditto_error and os.path.exists("/usr/bin/aa"):
             try:
                 unpack_cmd = [
                     "/usr/bin/aa",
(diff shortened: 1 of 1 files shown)


homebysix (approved):
Thanks!

PR #1051, opened 2026-06-05
Add `continue_on_error` input to PathDeleter

smithjw:
This PR adds an optional `continue_on_error` input to the `PathDeleter` processor (defaulting to `False`). When enabled, deletion becomes best-effort so an optional cleanup step (for example clearing a cache, removing build artifacts, etc) won't fail an otherwise-successful recipe run:

- a missing path is skipped instead of raising an error
- a directory that still can't be removed after retries is force-removed with `ignore_errors=True`

Directory removal also retries transient `OSError`s (for example a path briefly held open right after a build) three times with exponential backoff before giving up.

## Backward compatibility

The default (`continue_on_error=False`) keeps the existing behaviour of the Processor: any failure still raises a `ProcessorError`. The one change in default mode is the transient-failure retry on directory removal, which makes an eventual failure more robust rather than changing its outcome.

## Testing

Adds `Code/tests/test_pathdeleter.py` (9 tests): file, link, and directory happy paths, missing-path handling in both modes, and the retry and disposition logic (retry then succeed, retry then raise, retry then ignore). All pass under the bundled AutoPkg Python. black, isort, and flake8 are clean.

## Background

This behaviour is being pulled from my [SharedProcessor `FriendlyPathDeleter`](https://github.com/autopkg/smithjw-recipes/blob/master/SharedProcessors/FriendlyPathDeleter.py), using a clearer input name (`continue_on_error`). One deliberate difference from that original is that `continue_on_error` is respected for file and symlink removals too, not just directories and missing paths, so the flag behaves consistently across every path type.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -118,6 +118,7 @@ Multiple processors now confine paths to their intended directories, preventing 
 
 - Processors that handle disk images now preserve original mount failures instead of masking them with subsequent "not mounted" cleanup errors.
 - PkgCreator and AppPkgCreator: new `pkgbuild_args` input variable allows forwarding additional flags (e.g. `--filter`, `--large-payload`) to the `pkgbuild` tool (#981)
+- PathDeleter: new `continue_on_error` input variable (default `False`) makes deletion best-effort: a missing path is skipped instead of raising, and a directory that still can't be removed after a few retries is force-removed with errors ignored, so optional cleanup steps don't fail the recipe run. Directory removal now also retries transient failures (e.g. a path briefly held open after a build) with exponential backoff before giving up. The default behavior is unchanged: any failure still raises a `ProcessorError`.
 - macOS-only processors that rely on `hdiutil`, `pkgutil`, `xar`, or `pkgbuild` now fail with explicit platform errors on non-macOS instead of attempting to launch unavailable tools.
 - URLDownloaderPython now validates cached files against their actual size, exposes computed hashes, and preserves downloads when ETag or Last-Modified headers are missing.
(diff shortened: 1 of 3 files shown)


homebysix:
As long as we're touching PathDeleter, why not make continue the default behavior? I think this stoppage has been the subject of some criticism and has required workarounds in the past.

smithjw:
@homebysix The initial thought was to retain existing behaviour, but more than happy to have `continue_on_error` default to true

homebysix:
OK, I gathered some community consensus and they agree that your approach is better in order not to break existing behavior where recipe authors might expect a failure if a path doesn't exist.

I wonder whether the exponential backoff retry is overcomplicated for AutoPkg's use case, though. Have you actually encountered the "path briefly held open" issue? Would a simple branch like this solve the same issue more simply?

```py
elif os.path.isdir(path):
    try:
        shutil.rmtree(path)
    except OSError:
        if not continue_on_error:
            raise
        time.sleep(1)
        shutil.rmtree(path, ignore_errors=True)
```

Minor thing: could you add a unit test for symlink (`islink=True`) path?

smithjw:
I wrote the initial FriendlyPathDeleter many years ago and am pretty sure there were a couple times where that backoff helped in terms of deleting paths. I had that processor set to 5 retires, but dropped it to three for the commit to the core processor.

Best case, the user never sees a delay because it's deleted first time, worst case there's a minor delay before success/error if the first deletion fails.

Will add a test for the `islink=True` today.

smithjw:
@homebysix New tests added and conflicts fixed with this PR :)

homebysix:
Thanks!

PR #1052, opened 2026-06-06
Improved handling of bad/expired GitHub tokens

### LINKED ISSUE #1037: Evaluate adopting PyGithub-based GitHub API library from dev-3.x
## Background

The `dev-3.x` branch overhauled `autopkglib.github` into a new module called `autopkglib.apgithub`, rewriting it to use the [PyGithub](https://github.com/PyGithub/PyGithub) third-party library instead of our current curl/URLGetter-based implementation. This was never ported to `dev`.

Key changes: `GitHubSession` no longer inherits from `URLGetter`; release matching (prerelease vs. latest vs. regex) was improved; the module was renamed (breaking change for third-party processors that import `autopkglib.github` directly); and `PyGithub` was added as a dependency.

## Questions to resolve

1. Is the prerelease/latest/regex improvement worth adopting PyGithub, or can it be ported independently?
2. Is the breaking rename acceptable for 3.0? (Since it's a major version, this may be the right time.)

## Minimum action (independent of library decision)

- [x] Resolve the token path inconsistency in `dev`: `github/__init__.py` defaults to `~/.autopkg_gh_token` but `autopkglib/__init__.py` defines `DEFAULT_GH_TOKEN` pointing to `~/Library/AutoPkg/gh_token`. Pick one and use it consistently, with a fallback that reads the old path if the new one doesn't exist.

homebysix:
Users with a malformed or expired GitHub token currently see hard failures across all authenticated API calls. Recipe runs, repo operations, and searches can fail with no guidance on why. This change makes AutoPkg degrade gracefully to unauthenticated behavior and tell users what to do, so a stale token file doesn't silently break their setup.

**Token sanitization**

`get_github_token` now strips tokens and rejects ones that are empty or contain interior whitespace, logging a clear message and falling back to unauthenticated. A malformed preference token no longer silently falls through to the token file.

**401 retry**

If an authenticated GET returns 401, `call_api` logs a warning (with a link to regenerate the token) and retries once without auth. Non-GET requests get the warning but no retry, since anonymous writes/mutations don't make sense.

**Mechanics**

- Removed `--fail` from `GitHubSession`'s curl command so 4xx responses flow back as `(body, status)` instead of raising before status can be read. This also revives the existing 403/rate-limit handling path in `do_gh_repo_contents_fetch`, which was previously unreachable.
- Extracted `_call_api_once` to handle the single-request lifecycle, including a fix for a pre-existing temp file leak.
- Fixed a `KeyError` in `do_gh_repo_contents_fetch` where non-200 API error bodies (now returned instead of raised) could reach `results["content"]`.

**Validation**

In addition to the unit tests, which all pass, I used an AI agent to create a [one-off script](https://gist.github.com/homebysix/9474e41731b6a329bd475b44f9b9ed02) specifically to test the code paths that this PR changes. All pass:

```
Path 1: Malformed token (interior whitespace) is rejected
  [PASS] Returns None
  [PASS] Logs 'Ignoring malformed'
         stderr: Ignoring malformed GitHub token from /var/folders/5s/x9b8f6c11ln77h34hbwk_kzm0000gn/T/tmpo437lbf2/.autopkg_gh_token; continuing unauthenticated.
  [PASS] Names the source file
         stderr: Ignoring malformed GitHub token from /var/folders/5s/x9b8f6c11ln77h34hbwk_kzm0000gn/T/tmpo437lbf2/.autopkg_gh_token; continuing unauthenticated.

Path 2: Token without known prefix is accepted without warnings
  [PASS] Token is accepted
  [PASS] No warnings logged

Path 3: Unauthenticated session makes a successful API call
  (1 real network request to api.github.com)
  [PASS] HTTP 200 returned
         status: 200
  [PASS] Response contains expected repo name
         name field: autopkg
  [PASS] No token was sent (session.token is None)

Path 4: Bad token triggers 401, retries anonymously, succeeds
  (2 real network requests to api.github.com)
  stderr: WARNING: Your GitHub token appears invalid or expired. Regenerate it at https://github.com/settings/tokens. Continuing without it.
  [PASS] HTTP 200 returned after anonymous retry
         status: 200
  [PASS] Response contains expected repo name
         name field: autopkg
  [PASS] Warning: token invalid or expired
  [PASS] Warning: continuing without it

12 of 12 checks passed
```

I'm also running this today against a random sampling of AutoPkg in the org to check for regressions.

Closes #1037

### DIFF
--- a/Code/autopkg
+++ b/Code/autopkg
@@ -1075,7 +1075,7 @@ def do_gh_repo_contents_fetch(
             log_err(results.get("documentation_url", None))
 
         return None
-    if results is None or code is None:
+    if code is None or code != 200 or results is None:
         log_err("A GitHub API error occurred!")
         return None
     if decode:
--- a/Code/autopkgcmd/searchcmd.py
+++ b/Code/autopkgcmd/searchcmd.py
@@ -74,7 +74,7 @@ def check_search_cache(cache_path: str) -> None:
     # Use URLGetter to interact with GitHub API
     api = URLGetter()
 
-    # Use GitHub token if one exists
+    # GitHubSession filters malformed tokens; this path has its own fallback.
     token = GitHubSession().token
     headers = {"Authorization": f"Bearer {token}"} if token else {}
 
--- a/Code/autopkglib/github/__init__.py
+++ b/Code/autopkglib/github/__init__.py
@@ -30,20 +30,37 @@ TOKEN_LOCATION = os.path.expanduser("~/.autopkg_gh_token")
 DEFAULT_SEARCH_USER = "autopkg"
 
 
+def _sanitize_github_token(token: str | None, source: str) -> str | None:
+    """Return a usable GitHub token or None when it is clearly malformed."""
+    if token is None:
+        return None
+
+    token = token.strip()
+    if not token or re.search(r"\s", token):
+        log_err(
+            f"Ignoring malformed GitHub token from {source}; "
+            "continuing unauthenticated."
+        )
+        return None
+
+    return token
+
+
 def get_github_token(token_path: str | None = None) -> str | None:
     """Read token from preferences, then token_path."""
     token = get_pref("GITHUB_TOKEN")
     token_path = os.path.expanduser(token_path or TOKEN_LOCATION)
-    if not token and os.path.exists(token_path):
+    if token:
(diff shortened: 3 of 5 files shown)


homebysix:
No regressions flagged between this branch and dev branch.

PR #1056, opened 2026-07-11
Update `URLDownloader` & `URLDownloaderPython` with new inputs/validation

smithjw:
Aligns `URLDownloader` and `URLDownloaderPython` around shared cache metadata and change detection.

New inputs/outputs:
- `HEADERS_TO_TEST` selects the response headers used for change detection. Values are persisted in `.info.json`. Defaults to ETag, Last-Modified, and Content-Length.
- `download_missing_file` defaults to true and re-downloads a missing cached file without marking the remote version as changed. Set it to false for a metadata-only cache hit.
- `download_url` reports the final URL after redirects.
- `download_info` reports metadata from the previous or current download.

Also:
- Both processors now share case-insensitive header comparison, file-size fallback, metadata persistence, and summary reporting.
- Boolean inputs accept booleans and common string forms (`true`/`false`, `yes`/`no`, `on`/`off`, `1`/`0`). Null uses the declared default; a blank string is false. Invalid `prefetch_filename` values point users to `filename`.
- Missing-file refreshes rewrite `.info.json`, publish current metadata and hashes, and report the re-download.
- `CHECK_FILESIZE_ONLY` no longer mutates `HEADERS_TO_TEST` in the shared recipe environment.
- Malformed Content-Length values warn and are treated as changed rather than crashing.

Compatibility note: stricter boolean parsing exposes recipes that previously supplied arbitrary strings to boolean flags, and correctly treats the string `False` as false. Follow-up recipe fixes are open in autopkg/NickETH-recipes#11 and autopkg/flywire-recipes#24.

### DIFF
--- a/Code/autopkglib/URLDownloader.py
+++ b/Code/autopkglib/URLDownloader.py
@@ -109,8 +109,8 @@ class URLDownloader(URLGetter):
             "required": False,
             "description": (
                 "List of HTTP headers to compare against the previous download "
-                "to detect changes. If 'CHECK_FILESIZE_ONLY' is enabled, this "
-                "list is overridden to ['Content-Length'] only."
+                "to detect changes. Their values are persisted in .info.json. "
+                "If 'CHECK_FILESIZE_ONLY' is enabled, only Content-Length is used."
             ),
             "default": ["ETag", "Last-Modified", "Content-Length"],
         },
@@ -121,7 +121,8 @@ class URLDownloader(URLGetter):
                 "download the file again. Defaults to True as most current "
                 "recipes expect the files to be present. This re-fetch does "
                 "not mark the item as changed (download_changed stays false); "
-                "download_changed reflects the remote resource only."
+                "download_changed reflects the remote resource only. A missing "
+                "or null value uses the default; a blank string disables it."
             ),
             "default": True,
         },
@@ -164,7 +165,7 @@ class URLDownloader(URLGetter):
 
         value = self.env[key]
         if value is None:
-            return False
+            return default
         if isinstance(value, bool):
             return value
         if isinstance(value, str):
@@ -174,10 +175,13 @@ class URLDownloader(URLGetter):
             if normalised in ("false", "no", "off", "0", ""):
                 return False
 
-        raise ProcessorError(
+        message = (
             f"{key} must be a boolean or boolean-like string "
             f"(true/false, yes/no, on/off, 1/0), not {value!r}"
         )
+        if key == "prefetch_filename":
+            message += "; use filename to override the downloaded filename"
+        raise ProcessorError(message)
 
     def prepare_base_curl_cmd(self) -> list[str]:
         """Assemble base curl command and return it."""
@@ -221,7 +225,6 @@ class URLDownloader(URLGetter):
         # unchanged content.
         if os.path.exists(self.env["pathname"]):
             metadata = self.get_metadata()
-            self.existing_file_size: int | None = os.path.getsize(self.env["pathname"])
             if not self.env_bool("CHECK_FILESIZE_ONLY"):
                 http_headers: dict[str, Any] = metadata.get("http_headers", {})
                 if etag := http_headers.get("ETag"):
@@ -248,7 +251,6 @@ class URLDownloader(URLGetter):
         self.env["last_modified"] = ""
         self.env["etag"] = ""
         self.env["download_url"] = ""
-        self.existing_file_size = None
 
     def prefetch_filename(self) -> str | None:
         """Attempt to find filename in HTTP headers."""
@@ -371,10 +373,7 @@ class URLDownloader(URLGetter):
             return
         hash_keys = ("file_sha1", "file_sha256", "file_md5")
         if os.path.isfile(self.env["pathname"]):
-            hashes = self.compute_hashes()
-            self.env["file_sha1"] = hashes["sha1"]
-            self.env["file_sha256"] = hashes["sha256"]
-            self.env["file_md5"] = hashes["md5"]
+            self.store_hashes_in_env(self.compute_hashes())
             return
         metadata = self.env.get("download_info") or {}
         if all(metadata.get(key) for key in hash_keys):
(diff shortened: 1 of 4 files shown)


smithjw:
Reworked how `download_changed` and `download_missing_file` interact so change detection is about the remote resource, not whether the file happens to be present on disk.

**`download_changed`** now means one thing: does the remote differ from the cached `.info.json`? It no longer depends on the file being present.

**`download_missing_file`** only decides whether to re-fetch a file that has gone missing while the remote resource is unchanged. It never flips `download_changed`. Default stays `true`, so existing recipes keep getting the file.

| `.info.json` | Remote vs cached | Cached file | `download_missing_file` | `download_changed` | File (re)downloaded? |
|---|---|---|---|---|---|
| present | same | present | ignored | `false` | no (use cache) |
| present | same | absent | `true` | `false` | yes (re-fetched) |
| present | same | absent | `false` | `false` | no (skip) |
| present | differs | any | ignored | `true` | yes |
| absent | — | any | ignored | `true` | yes (first run) |

Also fixed a crash: with `COMPUTE_HASHES` enabled and the cached file missing, both processors tried to hash a file that wasn't there. Hashes now come from the file when present, from `.info.json` when not, and are skipped with a warning otherwise.

Applies to both `URLDownloader` and `URLDownloaderPython`.

homebysix:
1\. The new `env_bool()` is stricter than the old truthiness checks, and I think that's a good idea. I had a look through the recipes in the autopkg org for these four keys, and a couple of things turned up.

One recipe fails outright. NickETH-recipes' [Python-Win64 download recipe](https://github.com/autopkg/NickETH-recipes/blob/b363900c3e675a4fab49d14b0fd6616a8e6eb9d0/Python/Python-Win64.download.recipe.yaml#L20) passes `prefetch_filename: '%NAME%-x64.exe'`. It's a flag, so the string has been ignored all along — I think they meant `filename:`. After this PR that's an error on every run.

A few others quietly change behavior. Some recipes in NickETH-recipes and flywire-recipes pass `CHECK_FILESIZE_ONLY` as the string `'False'` ([TablePlus](https://github.com/autopkg/flywire-recipes/blob/78f9eb9123005b1c3aa77756bf8084538dc1b447/TablePlus/TablePlus.download.recipe#L11-L12) is one), which has meant "on" this whole time. Under this PR they'll start comparing ETag and Last-Modified again, so probably one extra re-import the first time they run. That seems like the right outcome, since it's what those recipes were always trying to say.

I've opened PRs against both repos: [NickETH-recipes#11](https://github.com/autopkg/NickETH-recipes/pull/11) and [flywire-recipes#24](https://github.com/autopkg/flywire-recipes/pull/24). For the Python one I just removed `prefetch_filename` — as far as I could tell it wasn't finding anything the URL wouldn't already give us.

Nothing else I looked at would break. Variable substitution converts values to strings before the processor sees them, so a `1` arrives as `"1"` and parses fine. I'd still soften "existing recipes keep working without edits" in the description. Might also be worth mentioning `filename:` in the error text, since that's the mistake someone's most likely making.

2\. `.info.json` persistence isn't lined up between the two when a file goes missing and gets re-fetched with the version unchanged. `URLDownloader` rewrites `.info.json` in that situation and `URLDownloaderPython` doesn't. `URLDownloaderPython` also doesn't log anything or set `url_downloader_summary_result` on that path, so it looks like nothing happened at all.

For those who run AutoPkg in CI, I can imagine some of them might want `.info.json` files whose hashes are still valid, to avoid re-fetching the download. I don't know of anybody doing that currently, but I imagine somebody probably does. Might be worth asking the community.

3\. The two `download_changed()` implementations disagree. Same metadata and the same response headers give two different answers:

```
HEADERS_TO_TEST: ['etag']           URLDownloader: unchanged   URLDownloaderPython: changed
HEADERS_TO_TEST: ['Last-Modified']  URLDownloader: unchanged   URLDownloaderPython: changed
```

Lowercase header names work in `URLDownloader` thanks to the new `metadata_header()`, and raise `KeyError` in `URLDownloaderPython`, which falls through to assuming the download changed. `URLDownloader`'s file-size fallback also isn't reachable from `URLDownloaderPython`, since `existing_file_size` only gets set in `produce_etag_headers()`.

Related: `URLDownloaderPython` writes `self.env["HEADERS_TO_TEST"] = ["Content-Length"]` when `CHECK_FILESIZE_ONLY` is set, where `URLDownloader` keeps it local. That write isn't new — it's been there since before 2.9.0, and other processors do the same kind of thing, [`MunkiImporter`](https://github.com/autopkg/autopkg/blob/cc240a6c8f2e12b1cdd82e908642abe7b15a5cca/Code/autopkglib/MunkiImporter.py#L466) rewrites `pkg_path` on purpose so later steps pick it up.

I did go looking for recipes that could hit it. `env` gets reset for each recipe in a run but carries across a parent chain, so the shape you'd need is two download steps in one chain where only the first sets `CHECK_FILESIZE_ONLY`. That does exist — [SketchUp2025](https://github.com/autopkg/markkenny-recipes/blob/master/Sketchup/SketchUp2025.download.recipe.yaml), [SketchUp 2025 DE](https://github.com/autopkg/jannheider-recipes/blob/main/SketchUp%202025%20DE/SketchUp%202025%20DE.download.recipe), [Localized-LibreOffice](https://github.com/autopkg/jpiel-recipes/blob/master/Localized-LibreOffice/Localized-LibreOffice.download.recipe) — but all three use `URLDownloader` for both steps, which keeps it local. Nothing pairs `URLDownloaderPython` with a later download step. Worth knowing this PR widens it slightly, since `URLDownloader` now reads `HEADERS_TO_TEST` out of the env where it didn't before, so a mixed chain would newly be affected. Nobody's doing that today, so I'll leave it up to you whether you want to align that behavior or not.

4\. `env_bool()` returns `False` for an explicit `None` before it looks at `default`, so a caller asking for `default=True` doesn't get it. I'd fix both sides of this. Documenting is the main part, since `input_variables` says `default: True` while a blank value means false and that isn't written down anywhere. But the guard is cheap too — returning `default` instead of `False` when the value is `None` is one line, and `""` still maps to false for the substitution case. There's a test asserting the current behavior that would need updating alongside it.

Nothing in the org depends on either behavior today. You can't get `None` in from a plist since there's no null type, and no recipe currently uses a bare YAML null. Substitution won't produce one either — `getdata()` turns `None` into `""`, which is already treated as false. The `DOWNLOAD_MISSING_FILE: null` entries in smithjw-recipes are the Input variable rather than the processor argument, and they substitute down to `""` → false, same as [that README](https://github.com/autopkg/smithjw-recipes/blob/ed32255111e97b40bb5541332c74cd60c9c41fbf/README.md) says.

smithjw:
Updated in 5bd6baf.

The behaviour described in my first comment still holds. `download_changed` only reflects the remote version. `download_missing_file` only controls whether a missing cached file is restored.

The follow-up review is also addressed:

- Missing-file refreshes now rewrite `.info.json`, log the action, and set the summary
- Both downloaders share the same case-insensitive header checks and file size fallback
- `CHECK_FILESIZE_ONLY` no longer changes `HEADERS_TO_TEST` in the shared env
- Null uses the declared boolean default; an empty string remains false
- Invalid `prefetch_filename` values now point to `filename`

I also removed the duplicate Python change detection and metadata code. The branch is rebased onto current `dev`, and all checks pass.

homebysix:
Looks good now. A few follow-ups I'll do after merging:

1\. `clear_vars()` resets `download_url` but not `download_info`, so it can carry over between processors in a run. I'll set it to `{}` alongside the others.

2\. I'd like to uppercase `download_missing_file` like the other input variables.

3\. `HEADERS_TO_TEST` persists the default headers on top of whatever you set but compares only what you set. That seems deliberate, since `store_metadata()` reads `ETag` and `Last-Modified` straight out of that dict. I'll just fix the description.

And two for you to consider:

4\. [Tableau_Prep_Builder.download.recipe.yaml](https://github.com/autopkg/smithjw-recipes/blob/master/Tableau/Tableau_Prep_Builder.download.recipe.yaml) passes `download_missing_file: '%DOWNLOAD_MISSING_FILE%'` without defining it in `Input`. Harmless now since the key is ignored, but once this PR merges the unsubstituted string reaches `env_bool` and the recipe will error.

5\. `DOWNLOAD_MISSING_FILE: null` in your download recipes (like [Ghostty](https://github.com/autopkg/smithjw-recipes/blob/master/Ghostty/Ghostty.download.recipe.yaml)) never gets to the processor as null. [`update_data()`](https://github.com/autopkg/autopkg/blob/06b9f17/Code/autopkglib/__init__.py) turns `None` into an empty string during substitution, and `env_bool` takes an empty string as false, so those recipes end up with false instead of the True default. Nothing breaks, but set them to `true`, or drop both the key and the argument, if you meant them to take the default.

PR #1061, opened 2026-08-02
GithubReleasesInfoProvider infinite loop fix

sboissez:
Fixes a small bug that causes the processor to loop forever if `latest_only: True` and no matching asset is found.

The bug happens because the page keeps incrementing up, but `latest_only` only checks the latest page. It keeps checking the latest page forever.

Can replicate the bug and fix using the following basic recipe:
```
Identifier: local.test.GithubLoop
Process:
    Arguments:
      github_repo: autopkg/autopkg
      latest_only: True
      asset_regex: "^this-will-never-match\\.pkg$"
```


homebysix:
Good catch — thank you!

PR #1066, opened 2026-09-03
Preserve URLDownloaderPython downloads when the server omits Content-Length

### LINKED ISSUE #1065: URLDownloaderPython discards the downloaded file when the server omits Content-Length
**Describe the problem**

`URLDownloaderPython` throws away a successfully downloaded file when the HTTP response has no `Content-Length` header (e.g. `Transfer-Encoding: chunked`). The bytes are streamed to a temp file, but that temp file is never moved to `pathname`, so whatever was previously at `pathname` is left in place — while the processor still reports `download_changed: True` and prints `Downloaded <pathname>`. Every downstream processor then operates on the stale file.

In `Code/autopkglib/URLDownloaderPython.py`, `download_and_hash()` builds the header dictionary *before* moving the temp file into place:

```python
try:
    # save http header info to dict
    download_dictionary["http_headers"] = {}
    download_dictionary["http_headers"]["Content-Length"] = int(
        response.headers["content-length"]
    )
    download_dictionary["http_headers"]["ETag"] = response.headers["ETag"]
    download_dictionary["http_headers"]["Last-Modified"] = response.headers[
        "Last-Modified"
    ]
    if download_dictionary["http_headers"]["Content-Length"] != size:
        # should this be a halting error?
        self.output("WARNING: file size != content-length header")
except (KeyError, TypeError) as err:
    # probably need to handle a missing header better than this
    self.output(
        "ERROR: header issue ({err_type})\n{err}\n".format(
            err=err, err_type=type(err).__name__
        )
    )
    return None

if self.env.get("download_changed", None):
    # Move the new temporary download file to the pathname
    self.move_temp_file(file_save_path)
```

On a chunked response `response.headers["content-length"]` is `None`, so `int(None)` raises `TypeError`. The handler returns `None` — and that early return sits *above* the `move_temp_file()` call. The download is already complete and correct at that point; it is simply abandoned.

`main()` then carries on as though it succeeded, because `self.env["download_changed"]` was set earlier in `download_and_hash()` and is never reset:

```python
if self.env.get("download_changed", None):
    # store download info for checking for existing download
    self.store_download_info_json(download_dictionary)   # download_dictionary is None
    self.output(f"Downloaded {self.env['pathname']}")     # not actually true
```

So the run reports a new download, writes an `info.json` containing the literal `null`, and leaves the real bytes in an orphaned `tmpXXXXXXXX` file. Because the `info.json` is `null`, the next run's `download_changed()` can never match cached headers either, so the recipe re-downloads and re-fails every time.

**Reproduction**

`ChunkedRepro.download.recipe.yaml`:

```yaml
Description: Minimal reproduction of URLDownloaderPython dropping a chunked download.
Identifier: com.github.example.download.ChunkedRepro
MinimumVersion: '2.9'

Input:
  NAME: ChunkedRepro
  DOWNLOAD_URL: https://www.synaptics.com/sites/default/files/exe_files/2026-07/DisplayLink%20Manager%20Graphics%20Connectivity16.2-EXE.pkg

Process:
  - Processor: URLDownloaderPython
    Arguments:
      filename: ChunkedRepro.pkg
      url: '%DOWNLOAD_URL%'
```

That host (Synaptics behind Cloudflare) sends no `Content-Length`:

```
$ curl -s -o /dev/null -D - "https://www.synaptics.com/sites/default/files/exe_files/2026-07/DisplayLink%20Manager%20Graphics%20Connectivity16.2-EXE.pkg"
HTTP/2 200
content-type: text/plain
transfer-encoding: chunked
etag: W/"6a5f37d4-9eabf8"
last-modified: Tue, 21 Jul 2026 09:11:48 GMT
server: cloudflare
```

Seed the cache with a stand-in for a previously downloaded version, then run:

```
$ C="$HOME/Library/AutoPkg/Cache/com.github.example.download.ChunkedRepro"
$ mkdir -p "$C/downloads"
$ printf 'PREVIOUS VERSION PLACEHOLDER\n' > "$C/downloads/ChunkedRepro.pkg"
$ autopkg run -vv ChunkedRepro.download.recipe.yaml --search-dir .
```

**AutoPkg output**

```
URLDownloaderPython: WARNING: missing download info (FileNotFoundError)
URLDownloaderPython: WARNING: 'Content-Length' missing. (TypeError) 'NoneType' object is not subscriptable
URLDownloaderPython: WARNING: header missing. (TypeError) 'NoneType' object is not subscriptable
URLDownloaderPython: WARNING: header missing. (TypeError) 'NoneType' object is not subscriptable
URLDownloaderPython: ERROR: header issue (TypeError)
URLDownloaderPython: Downloaded .../downloads/ChunkedRepro.pkg
    Download Path
```

Cache directory afterwards:

```
      29  ChunkedRepro.pkg             <- the stand-in, untouched
       5  ChunkedRepro.pkg.info.json   <- contains "null"
10398712  tmp_4o9a_q7                  <- the real download, orphaned
```

The stock curl-based `URLDownloader` handles the same URL correctly and writes the 10398712-byte package to `pathname`, so this is specific to `URLDownloaderPython`.

**Expected behavior**

A download that completed successfully should be moved to `pathname`. A missing optional response header should be recorded as absent (and ideally warned about), not treated as a reason to discard the file. Conversely, if the download really is unusable, `download_changed` should be `False` and the processor should not report `Downloaded ...`.

**Impact**

`Content-Length` is optional in HTTP/1.1 and HTTP/2, and CDNs routinely omit it on chunked responses, so this isn't an exotic edge case. Two ways it bites:

1. **Silent staleness.** With a previous version already cached, `CodeSignatureVerifier`, `FlatPkgUnpacker`, `MunkiImporter` and friends all run against the *old* artefact while the run reports a new download.
2. **Misleading failures.** In an ephemeral/cloud CI setup the file at `pathname` may be a size-accurate placeholder rather than a real package. `CodeSignatureVerifier` then fails with `Code signature verification failed`, which sends you investigating certificates when the real problem is that the download never landed. Following the error's own advice and setting `DISABLE_CODE_SIGNATURE_VERIFICATION` just moves the failure to the unpacker.

There is no recipe-level workaround via `request_headers`: `Accept-Encoding: gzip` does yield a `Content-Length`, but the body then arrives with `content-encoding: gzip` and `urlopen()` doesn't transparently decode it, so the saved file is corrupt; `Range: bytes=0-` returns `206` with a `content-range` but still no `content-length`.

**Suggested fix**

Treat the response headers as best-effort metadata rather than a precondition — record whichever validators the server actually sent, and always move the temp file into place when the download succeeded:

```python
# Record whichever validators the server actually sent. A missing header
# (Content-Length is absent on chunked responses) must not discard a
# download that has already completed successfully.
download_dictionary["http_headers"] = {}
for header in ("Content-Length", "ETag", "Last-Modified"):
    value = response.headers.get(header)
    if value is None:
        self.output(f"WARNING: '{header}' missing from response headers", 1)
        continue
    download_dictionary["http_headers"][header] = (
        int(value) if header == "Content-Length" else value
    )

content_length = download_dictionary["http_headers"].get("Content-Length")
if content_length is not None and content_length != size:
    self.output("WARNING: file size != content-length header")

if self.env.get("download_changed", None):
    self.move_temp_file(file_save_path)
```

`download_changed()` already tolerates missing headers — it skips any that raise `KeyError`/`TypeError` and falls back to `return True` when nothing could be compared — so a partial `http_headers` dict degrades gracefully to "always re-download" rather than breaking caching.

Happy to open a PR with this.

**Version**

- OS version: macOS 26.2 (also reproduces on Bitrise's macOS CI runners)
- AutoPkg Version: 2.9.0

adambutterwick-roo:
Fixes #1065.

## What

`URLDownloaderPython.download_and_hash()` treated `Content-Length`, `ETag` and `Last-Modified` as required response headers. All three are optional, and CDNs routinely omit `Content-Length` on chunked responses, in which case `int(response.headers["content-length"])` raised `TypeError`. The handler for that returned `None` — and the early return sits *above* the `self.move_temp_file(file_save_path)` call, so a download that had already completed successfully was abandoned in its temp file.

Because `self.env["download_changed"]` was set earlier and never reset, `main()` then carried on as though the download had worked: it wrote an `info.json` containing the literal `null` and printed `Downloaded <pathname>`, while `pathname` still held the previous run's artefact. Downstream processors consumed the stale file, and the `null` info.json meant the next run could never match cached headers, so the recipe re-downloaded and re-failed every time.

This changes the header handling to best-effort: record whichever validators the server actually sent, warn about the ones it didn't, and always move the temp file into place when the download succeeded. The `file size != content-length header` warning is kept, but only when a `Content-Length` was actually supplied.

`download_changed()` already tolerates missing headers — it skips any that raise `KeyError`/`TypeError` and falls back to `return True` when nothing could be compared — so a partial `http_headers` dict degrades gracefully to "always re-download" rather than breaking caching.

## Test plan

Added `Code/tests/test_urldownloaderpython.py` with three cases covering a response with all validators, one without `Content-Length` (chunked), and one with no validators at all. Each asserts the download is moved into place and the recorded headers match what the server sent. The two new-behaviour cases fail against the unpatched processor with `Expected 'move_temp_file' to have been called once. Called 0 times.`

- `python -m unittest discover -s tests -p "test_*.py"` → 742 tests pass
- `black`, `isort`, `flake8` clean on both changed files

Verified end-to-end against a live chunked host (Synaptics behind Cloudflare, `transfer-encoding: chunked`, no `content-length`), with a stand-in for a previously downloaded version already at `pathname`.

Before:

```
URLDownloaderPython: ERROR: header issue (TypeError)
URLDownloaderPython: Downloaded .../downloads/ChunkedRepro.pkg

      29  ChunkedRepro.pkg             <- the stand-in, untouched
       5  ChunkedRepro.pkg.info.json   <- contains "null"
10398712  tmp_4o9a_q7                  <- the real download, orphaned
```

After:

```
URLDownloaderPython: WARNING: 'Content-Length' missing from response headers
URLDownloaderPython: Downloaded .../downloads/ChunkedRepro.pkg

10398712  ChunkedRepro.pkg
     341  ChunkedRepro.pkg.info.json
```

`pkgutil --check-signature` on the resulting package passes, and the recorded metadata keeps caching working:

```json
{
    "file_name": "ChunkedRepro.pkg",
    "file_size": 10398712,
    "http_headers": {
        "ETag": "W/\"6a5f37d4-9eabf8\"",
        "Last-Modified": "Tue, 21 Jul 2026 09:11:48 GMT"
    }
}
```

A second run against the unchanged URL correctly short-circuits with `Nothing downloaded, packaged or imported.` and leaves no temp files behind, so `StopProcessingIf download_changed == False` still behaves as intended.

Made with Cursor

### DIFF
--- a/Code/autopkglib/URLDownloaderPython.py
+++ b/Code/autopkglib/URLDownloaderPython.py
@@ -334,28 +334,23 @@ class URLDownloaderPython(URLDownloader):
             download_dictionary["file_sha256"] = hashes[1].hexdigest()
             download_dictionary["file_md5"] = hashes[2].hexdigest()
         download_dictionary["download_url"] = url
-        # download_dictionary['http_headers'] = response.info()
-        try:
-            # save http header info to dict
-            download_dictionary["http_headers"] = {}
-            download_dictionary["http_headers"]["Content-Length"] = int(
-                response.headers["content-length"]
+        # Record whichever validators the server actually sent. All three are
+        # optional: Content-Length is absent on chunked responses, and a
+        # missing header must not discard a download that already succeeded.
+        download_dictionary["http_headers"] = {}
+        for header in ("Content-Length", "ETag", "Last-Modified"):
+            value = response.headers.get(header)
+            if value is None:
+                self.output(f"WARNING: '{header}' missing from response headers", 1)
+                continue
+            download_dictionary["http_headers"][header] = (
+                int(value) if header == "Content-Length" else value
             )
-            download_dictionary["http_headers"]["ETag"] = response.headers["ETag"]
-            download_dictionary["http_headers"]["Last-Modified"] = response.headers[
-                "Last-Modified"
-            ]
-            if download_dictionary["http_headers"]["Content-Length"] != size:
-                # should this be a halting error?
-                self.output("WARNING: file size != content-length header")
-        except (KeyError, TypeError) as err:
-            # probably need to handle a missing header better than this
-            self.output(
-                "ERROR: header issue ({err_type})\n{err}\n".format(
-                    err=err, err_type=type(err).__name__
-                )
-            )
-            return None
+
+        content_length = download_dictionary["http_headers"].get("Content-Length")
+        if content_length is not None and content_length != size:
+            # should this be a halting error?
+            self.output("WARNING: file size != content-length header")
 
         if self.env.get("download_changed", None):
             # Move the new temporary download file to the pathname
--- a/Code/tests/test_urldownloaderpython.py
+++ b/Code/tests/test_urldownloaderpython.py
@@ -1,0 +1,139 @@
+#!/usr/local/autopkg/python
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import email.message
+import os
+import shutil
+import sys
+import tempfile
+import unittest
+from unittest.mock import patch
+
+from autopkglib.URLDownloaderPython import URLDownloaderPython
+
+# autopkglib re-exports the class under the module's own name, so patch the
+# module object directly rather than by dotted path.
+URLDOWNLOADERPYTHON = sys.modules[URLDownloaderPython.__module__]
+
+
+def build_response(payload, headers):
+    """Return a stand-in for the object urlopen() hands back."""
+    message = email.message.Message()
+    for key, value in headers.items():
+        message[key] = value
+
+    class FakeResponse:
+        """Minimal urlopen() response exposing read() and headers."""
+
+        def __init__(self):
+            self._chunks = [payload, b""]
+
+        def read(self, _size=None):
+            return self._chunks.pop(0)
+
+        @property
+        def headers(self):
+            return message
+
+        def info(self):
+            return message
+
+    return FakeResponse()
+
+
+class TestURLDownloaderPython(unittest.TestCase):
+    """Test class for URLDownloaderPython Processor."""
+
+    def setUp(self):
+        """Set up test fixtures."""
+        self.temp_dir = tempfile.mkdtemp()
+        self.payload = b"downloaded package contents"
+        self.processor = URLDownloaderPython()
+        self.processor.env = {
+            "url": "http://example.com/file.pkg",
+            "filename": "file.pkg",
+            "RECIPE_CACHE_DIR": self.temp_dir,
+            "download_dir": self.temp_dir,
+            "pathname": os.path.join(self.temp_dir, "file.pkg"),
+        }
+
+    def tearDown(self):
+        """Clean up after tests."""
+        shutil.rmtree(self.temp_dir, ignore_errors=True)
+
+    def download(self, headers):
+        """Run download_and_hash() against a response with the given headers."""
+        temp_file = os.path.join(self.temp_dir, "tempfile")
+        with patch.object(URLDOWNLOADERPYTHON, "urlopen") as mock_urlopen, patch.object(
+            self.processor, "download_changed", return_value=True
+        ), patch.object(self.processor, "store_headers"), patch.object(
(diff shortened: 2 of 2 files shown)


adambutterwick-roo:
Worth flagging before this gets reviewed: this is already fixed on `dev` by eb9522f339dd8ee20791e93ebf626006943b78ee, so this PR only earns its keep if another 2.x release ships off `master` before 3.0. It also adds `Code/tests/test_urldownloaderpython.py`, which `dev` already has from that commit, so merging would create a conflict on the next `master` → `dev` merge.

Happy to close it — see https://github.com/autopkg/autopkg/issues/1065#issuecomment-5568050342 for the details and an offer to port the one missing test case to `dev` instead.

PR #1068, opened 2026-09-11
Fix GITHUB_RELEASES_PER_PAGE input being ignored in GitHubReleasesInfoProvider

vincentschall:
## Summary
The `GITHUB_RELEASES_PER_PAGE` input variable is defined and documented but
has no effect. In `main()`, releases are fetched with
`self.env.get("per_page", 30)`, which reads an undeclared key that is never
set, so per-page is always the hardcoded default of 30. This wires the fetch
to the documented input name.

## Test Plan
- Ran the full suite under AutoPkg's bundled Python
  (`/usr/local/autopkg/python Scripts/run_tests.py`): `Ran 739 tests ... OK
  (skipped=12)`, including all 6 `test_githubreleasesinfoprovider` tests. No
  regression.
- `black` / `isort` / `flake8` clean on the changed file.


vincentschall:
Saw that it is already fixed in dev in [f00e365](https://github.com/autopkg/autopkg/commit/f00e3654d7f2ea81d2153c54365b9482bd719e7c)

Issue #1063, opened 2026-08-26
hdiutil error and forthcoming deprecation in macOS 27


jelockwood:
**Describe the problem**
Whilst I normally trigger running recipes via AutoPkgr, I believe this error is more likely to be down to the underlying autopkg tool rather than AutoPkgr. This has been confirmed by running the same individual recipe via both AutoPkgr and AutoPkg.

I have previously seen occasional errors reported by AutoPkgr (really the error is coming from autopkg and merely being displayed by AutoPkgr) where it failed to unmount a disk image via the hdiutil command. More recently it is now producing _multiple_ such errors. I believe the fact this is now much more common is likely down to recent macOS 26 updates including 26.6.1 and 26.6.2.

Again previously this was worked-around by running the GUI disk utility app and ejecting these otherwise invisible but still mounted disk images with a subsequent run previously succeeding - meaning it was previously a more intermittent issue. It has now however become a much more inconvenient issue as the manual unmounting no longer prevents subsequent occurrences and it also seems more like _all_ disk image provided installers are now triggering this. As a result I now get during a full overnight run a huge list of these.

The hdiutil command is still part of macOS 26.6.2 but I am seeing web articles indicating it will be officially deprecated in macOS 27 aka Golden Gate. Therefore regardless of the current issue I am encountering and describing here, it would seem extremely urgent to replace the use of hdiutil with some alternative solution.

**AutoPkg output**
autopkg run local.munki.Poly\ Studio
Processing local.munki.Poly Studio...
unmounting /Users/munkiadmin/Library/AutoPkg/Cache/local.munki.Poly Studio/downloads/PolyStudio-5.1.0.1803.dmg failed: hdiutil: couldn't unmount "disk9" - Resource busy

Failed.

The following recipes failed:
    local.munki.Poly Studio
        Error in local.munki.Poly Studio: Processor: FlatPkgUnpacker: Error: unmounting /Users/munkiadmin/Library/AutoPkg/Cache/local.munki.Poly Studio/downloads/PolyStudio-5.1.0.1803.dmg failed: hdiutil: couldn't unmount "disk9" - Resource busy

Nothing downloaded, packaged or imported.

**Expected behavior**
Recipes downloading disk image based installers/updates to run without error - that is to download, mount, process, and then finally unmount the disk image.

**Version (please complete the following information):**
 - OS version: 26.6.2
- AutoPkg Version: 2.9.0

homebysix:
[...]

PR #1069, opened 2026-09-15
Unmount with diskutil eject

arussell:
Following on from #1063, under macOS 27, getting I'm getting the following error for every piece of software that is packaged in a `.dmg` file:

```
unmounting [...].dmg failed: hdiutil: WARNING: 'hdiutil detach ...' is deprecated. Please use 'diskutil eject ...' instead.
hdiutil: couldn't unmount "disk18" - Resource busy
```

This happens 100% of the time. Running `hdiutil detach` manually fails to eject the disk but `diskutil eject` works without issue.

This patch replaces only the `hdiutil detach` command, keeping `hdiutil` in place for all other tasks. I've also added the `-force` switch which helps with situations where security scanners are not done scanning the disk image by the time the eject needs to happen, reducing failure risk.

This still works without the `-force` switch, but adding it makes things a little more robust for some users.


Code/autopkglib/DmgMounter.py
```diff
@@ -164,15 +164,15 @@ def unmount(self, pathname) -> None:
         # Call hdiutil.
         try:
             proc = subprocess.Popen(
-                ("/usr/bin/hdiutil", "detach", self.mounts[pathname]),
+                ("/usr/sbin/diskutil", "eject", self.mounts[pathname], "-force"),
```
gregneagle, line 167: `-force` is a big change. `hdiutil detach` also supports `-force` and yet we don't use that, as the need for `-force` masks other issues.

I suspect your original issue would have also been addressed by adding `-force` to the `hdiutil detach` call. The question for you is: what else is accessing the disk image, preventing it from unmounting? (Often the answer is some sort of antivirus/antimalware/security software)

How far back does `diskutil eject` work with disk images? IOW, have you tested on older macOSes?

gregneagle:
If we're going to add `-force`, I'd rather it tried first without `-force`, then after failure without `-force`, it tried again with `-force`. 
```
Force will force-unmount the volumes (less kind to any open files; see also umount (8))
```
Maybe we don't care too much; force-unmounting a read-only volume is very unlikely to damage any files; but in the rare case of a read/write disk image it seems possible at least.

gregneagle:
Additionally, the documented/supported syntax is just `force` with no leading hyphen:
```
$ diskutil eject
'Usage:  diskutil eject [force] MountPoint|DiskIdentifier|DeviceNode
```

gregneagle:
Answering my own Q about how far back `diskutil eject` is supported -- at least as far back as 10.15:
```
$ diskutil eject
Usage:  diskutil eject MountPoint|DiskIdentifier|DeviceNode
$ sw_vers
ProductName:	Mac OS X
ProductVersion:	10.15.8
BuildVersion:	19H2036
$ diskutil eject force foo
Unable to find disk for force
$ diskutil eject -force foo
Unable to find disk for -force
```
But no `force` option it seems in 10.15. I'll leave it to others to discover when `force` was added.

gregneagle:
Let me speak a bit about the `hdiutil` deprecation issue. Yes, in macOS Tahoe, some `hdiutil` operations will cause deprecation warnings to be printed to stderr. But all `hdiutil` operations still work as they do in macOS Tahoe: nothing has yet been removed. There are some significant functionality gaps with the `diskutil image` suggested replacements. It will be important to file feedback with Apple so if they are going to truly remove (or cripple) `hdiutil`, that they make sure `diskutil` can be a complete replacement _before_ they remove `hdiutil`. There's no emergency here, but it's also something we can't ignore, either.

arussell:
The problem this is intended to solve is that `hdiutil detach` was returning `Resource busy` errors. I found that `hdiutil detach -force` did not solve this and still got `Resource busy` errors. This meant that `hdiutil` was incapable of unmounting the disk image regardless of the way it was called. I was also encountering this problem under macOS 26.6 (and possibly earlier releases) too, so this isn't specifically a new issue on macOS 27.

Next I tried `diskutil unmountDisk` which did not return any errors, but also silently failed causing the system to end up with a huge number of mounted disk images over time. This led me to `diskutil eject` which worked well in these situations, even without `force`.

The rationale for adding `force` was that I saw a couple of edge cases where the output of `diskutil eject` explicitly calls out a process that's locking the disk. This is quite useful as it tells you the culprit, for example:

```
Unmount of disk13 failed: at least one volume could not be unmounted
Unmount was dissented by PID 2777 (/usr/libexec/diskimagesiod)
Dissenter parent PPID 1 (/sbin/launchd)
```

 If this happens `force` will eject anyway in these situations -- similar to `Force Eject` in Finder, which I suspect uses the same underlying mechanism.

Your advice to only use `force` on a second attempt seems wise, no point forcing unnecessarily. Or if you wanted to not even do this, one potential alternative might be to only use `force` if an extra flag was passed to `autopkg run` expressly permitting use of `force` during eject, but that seems like a more complex change and I didn't want to overcomplicate this changeset by doing this right out of the gate. Happy to follow your preference here though.

homebysix:
@arussell I'd encourage you to test on the `dev` branch, which includes some fixes targeted at the "Resource busy" error. I do plan to release at least one more 3.0 RC before a final 3.0 release, which will include those fixes.

arussell:
@homebysix I tried the `dev` branch but unfortunately that did not solve or handle the `Resource busy` issues, which led to `autopkg run` skipping lots of updates and lots of disk images remaining mounted.

